Issue #1, opened 2018-11-27
Implement serialization and deserialization of `bool`


jannikluhn:
Implement serialization and deserialization of

- [ ] [bool](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#bool)

In order to be able to easily swap RLP with SSZ, we should try to expose the same API interface to the outside world, in particular:

- [`ssz.encode(obj, sedes=None)`](https://github.com/ethereum/pyrlp/blob/master/rlp/codec.py#L20)
- [`ssz.decode(obj, sedes)`](https://github.com/ethereum/pyrlp/blob/master/rlp/codec.py#L209)

The code should be easily and cleanly extendable to support other types.

[py-rlp](https://github.com/ethereum/pyrlp/) may be used as a reference. They split up encoding in two parts: First, they `serialize` arbitrary Python types using [sedes objects](https://github.com/ethereum/pyrlp/tree/master/rlp/sedes) to (nested lists of) bytes, which they `encode` in the second step to the final single byte string. The appropriate sedes type can either be passed explicitly or be inferred from the type of the Python object (note that the latter might not be possible in SSZ due to different encoding of fixed and arbitrary length byte objects). Deserialization works the other way around, but the sedes object must be passed explicitly, as there is no way to infer it. 

Sedes objects are also used to define composite types. While not part of this milestone, an [equivalent](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#container)  exists in SSZ and we'll need them later. It might be a good idea to stick to this design, but it is not necessary if we can come up with something better that also allows for defining composite types.

The code should be well tested. If an architecture similar to the one from `pyrlp` is chosen, the serialization/deserialization should be tested independently from encoding/decoding. There should be tests that check that

- valid inputs are serialized/deserialized/encoded/decoded to the correct outputs
- serialization/deserialization/encoding/decoding of invalid inputs is detected whenever possible and appropriate exceptions are thrown

If there are non-trivial parts of the code that are not directly touched by above tests, there should be separate test cases for those as well.

All tests should pass of course.

Bhargavasomu:
@jannikluhn I would like to take up this issue as this sounds interesting and I feel that I can do this. Do I have the permission to work on this?

cburgdorf:
@Bhargavasomu yes, feel free to start working on this.

jannikluhn:
[...]

PR #3, opened 2018-12-01
Add basic API infrastructure with Boolean Support

Bhargavasomu:
## What was wrong?
SSZ needs to implemented by starting with the infrastructure. This module needs to follows this[ Spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md)
Issue #1

## How was it fixed?
The infrastructure was mostly mimiced from `rlp`. The tests for `boolean serialization` and `deserialization` have been included.

#### Cute Animal Picture

image

### DIFF
--- a/ssz/codec.py
+++ b/ssz/codec.py
@@ -1,0 +1,29 @@
+from eth_utils import (
+    is_bytes,
+)
+
+from ssz.exceptions import (
+    DecodingError,
+)
+from ssz.utils import (
+    infer_sedes,
+)
+
+
+def encode(obj):
+    """
+    Encode object in SSZ format.
+    """
+    serialized_obj = infer_sedes(obj).serialize(obj)
+    return serialized_obj
+
+
+def decode(ssz, sedes):
+    """
+    Decode a SSZ encoded object.
+    """
+    if not is_bytes(ssz):
+        raise DecodingError('Can only decode SSZ bytes, got type %s' % type(ssz).__name__, ssz)
+
+    obj = sedes.deserialize(ssz)
+    return obj
--- a/ssz/utils.py
+++ b/ssz/utils.py
@@ -1,0 +1,23 @@
+from ssz.sedes import (
+    boolean,
+)
+
+
+def is_sedes(obj):
+    """
+    Check if `obj` is a sedes object.
+    A sedes object is characterized by having the methods
+    `serialize(obj)` and `deserialize(serial)`.
+    """
+    return hasattr(obj, 'serialize') and hasattr(obj, 'deserialize')
+
+
+def infer_sedes(obj):
+    """
+    Try to find a sedes objects suitable for a given Python object.
+    """
+    if isinstance(obj, bool):
+        return boolean
+
+    msg = 'Did not find sedes handling type {}'.format(type(obj).__name__)
+    raise TypeError(msg)
--- a/tests/core/test_boolean_serializer.py
+++ b/tests/core/test_boolean_serializer.py
@@ -1,0 +1,78 @@
+import pytest
+
+from ssz import (
+    DeserializationError,
+    SerializationError,
+)
+from ssz.sedes import (
+    Boolean,
+)
+
+
+@pytest.mark.parametrize(
+    'value,expected',
+    (
+        (True, b'\x01'),
+        (False, b'\x00'),
+    ),
+)
+def test_boolean_serialize_values(value, expected):
+    sedes = Boolean()
+    assert sedes.serialize(value) == expected
+
+
+@pytest.mark.parametrize(
+    'value',
+    (
+        None,
+        1,
+        0,
+        'True',
+        b'True',
+    ),
+)
+def test_boolean_serialize_bad_values(value):
+    sedes = Boolean()
+    with pytest.raises(SerializationError):
+        sedes.serialize(value)
+
+
+@pytest.mark.parametrize(
+    'value,expected',
+    (
+        (b'\x01', True),
+        (b'\x00', False),
+    ),
+)
+def test_boolean_deserialization(value, expected):
+    sedes = Boolean()
+    assert sedes.deserialize(value) == expected
+
+
+@pytest.mark.parametrize(
+    'value',
+    (
+        b' ',
+        b'\x02',
+        b'\x00\x00',
+        b'\x01\x00',
+        b'\x00\x01',
+        b'\x01\x01',
+    ),
+)
+def test_boolean_deserialization_bad_value(value):
+    sedes = Boolean()
+    with pytest.raises(DeserializationError):
+        sedes.deserialize(value)
+
+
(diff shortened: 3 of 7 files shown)


Bhargavasomu:
How come the `circleci` checks aren't running?

Bhargavasomu:
@jannikluhn could you take a look at the infrastructure built and could you also help me out with the `circleci`? `CircleCI` checks are not running for some reason.

hwwhww:
@Bhargavasomu I set up the CircleCI. Could you open a new PR to trigger it? Thanks.

Bhargavasomu:
@hwwhww @jannikluhn I have made some changes and force pushed, so that now it triggers `CircleCI`. Please give the review and I will make the changes likewise.

ssz/codec.py
```diff
@@ -0,0 +1,37 @@
...
+    DecodingError,
+)
+from ssz.utils import (
+    infer_sedes,
+)
+
+
+@lru_cache(128)
+def encode(obj):
+    '''
```
jannikluhn: Can we replace all `'''`s with `"""`s? This seems to be a convention in Ethereum's Python codebase, even if it isn't catched by the linter and I can't find a style guide requiring it.

hwwhww:
@Bhargavasomu
Due to the CI setting, the testing files should be under `tests/core/` directory, i.e., `tests/core/test_boolean_serializer.py`

ssz/codec.py
```diff
@@ -0,0 +1,37 @@
...
+
+from ssz.exceptions import (
+    DecodingError,
+)
+from ssz.utils import (
+    infer_sedes,
+)
+
+
+@lru_cache(128)
```
jannikluhn: I'm not convinced that we need this optimization and would rather remove it until we're sure that it actually makes a difference.

ssz/codec.py
```diff
@@ -0,0 +1,37 @@
...
+def encode(obj):
+    '''
+    Encode object in SSZ format
```
jannikluhn:
```suggestion
    Encode object in SSZ format.
```

ssz/codec.py
```diff
@@ -0,0 +1,37 @@
...
+    '''
+    serialized_obj = infer_sedes(obj).serialize(obj)
+    return serialized_obj
+
+
+@lru_cache(128)
```
jannikluhn: See above

ssz/codec.py
```diff
@@ -0,0 +1,37 @@
...
+    '''
+    Decode a SSZ encoded object
+    '''
+    if not is_bytes(ssz):
+        raise DecodingError('Can only decode SSZ bytes, got type %s' % type(ssz).__name__, ssz)
+    if sedes:
+        obj = sedes.deserialize(ssz)
+        return obj
+
+    return ssz
```
jannikluhn: This looks weird and might be a bug, can you explain what's happening here? If we keep it like this, it would probably be a good idea to add a comment.

What do you think about making `sedes` required to eliminate this code path altogether. Would that have any negative side effects I'm missing?

ssz/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    `serialize(obj)` and `deserialize(serial)`.
+    """
+    return hasattr(obj, 'serialize') and hasattr(obj, 'deserialize')
+
+
+def infer_sedes(obj):
+    '''
+    Try to find a sedes objects suitable for a given Python object.
+    '''
+    if is_sedes(obj.__class__):
```
jannikluhn: This is something we only get once we have containers, so I'd remove it for the time being.

ssz/codec.py
Bhargavasomu: This codepath is because of the behaviour of `rlp` I was trying to mimic. Consider the following
```
In : rlp.decode(b'\x05')
Out : b'\x05'
```
Here in `rlp` they are not making `sedes` compulsory and are returning what was given if `sedes` was not mentioned. But I don't see any meaning in that, and maybe we could make `sedes` compulsory.

Bhargavasomu: I will be modifying the code, so that an `exception` is raised if `sedes` object is not specified while `deserialization`. Is that ok @jannikluhn

Bhargavasomu:
@jannikluhn @hwwhww thankyou for the review. Made the necessary changes. Could you take a look again?

jannikluhn:
LGTM. Just one request: In the future, please don't force push to preserve the commit history and make reviewing easier.

Bhargavasomu:
@jannikluhn sorry about that, will avoid force push in the future. So if this is over and can be merged, what should my next step be in regards to the development of `ssz`

ssz/codec.py
```diff
@@ -0,0 +1,29 @@
...
+
+from ssz.exceptions import (
+    DecodingError,
+)
+from ssz.utils import (
+    infer_sedes,
+)
+
+
+def encode(obj):
```
hwwhww: Could you write a test for it to verify that it's callable?

ssz/codec.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Encode object in SSZ format.
+    """
+    serialized_obj = infer_sedes(obj).serialize(obj)
+    return serialized_obj
+
+
+def decode(ssz, sedes):
```
hwwhww: ditto

Bhargavasomu:
@hwwhww I have added the corresponding test case in `test_boolean_serializer.py` as the function `test_boolean_round_trip_codec`. Could you please check it?

ssz/utils.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+
+def is_sedes(obj):
+    """
+    Check if `obj` is a sedes object.
+    A sedes object is characterized by having the methods
+    `serialize(obj)` and `deserialize(serial)`.
+    """
+    return hasattr(obj, 'serialize') and hasattr(obj, 'deserialize')
```
pipermerriam: What do you all think about implementing and dectecting a `sedes` using a common base class (this approach bothers me because it is imprecise).

```python
class BaseSedes(ABC):
    @abstractmethod
    def serialize(self, obj):
        pass

    @abstractmethod
    def deserialize(self, serialized_obj):
        pass
```

This way we can just do an `isinstance` check instead of `hasattr` checks.  Thoughts?

tests/core/test_boolean_serializer.py
```diff
@@ -0,0 +1,92 @@
...
+)
+def test_boolean_deserialization(value, expected):
+    sedes = Boolean()
+    assert sedes.deserialize(value) == expected
+
+
+@pytest.mark.parametrize(
+    'value',
+    (
+        b' ',
```
pipermerriam: It'd be nice to add empty string to this list as well as the text strings `'\x00'` an `'\x01'` (as opposed to the bytestrings).

Issue #4, opened 2018-12-02
Implement serialization and deserialization of remaining base types


jannikluhn:
Implement serialization and deserialization of 

- [x] [uint](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#uint)
- [x] [address](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#address)
- [x] [hashN](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#hash)
- [x] [bytes](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#bytes)
- [x] [lists](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#listvectors)

~as well as~

- [ ] ~[tree hashes](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash) (excluding containers for now)~.

As always, the code should be well tested.

This should be quite straightforward. For the various `uint` and `hashN`/`bytes`/(maybe) `address` types, I suggest implementing a single sedes class for the base type that takes the size as an argument, but in addition to that to also export various instances for typical sizes (e.g., `uint64`, `hash32` and so on) to make using those as easy as possible.

Bhargavasomu:
On this

Bhargavasomu:
@jannikluhn while deserializing `intN` or `unitN` (in other sedes objects too, though not explicitly mentioned) is it necessary to return both the `deserialized object` and the `new_index`? Wouldn't returning only the `deserialized_object` be sufficient. In other words I didn't understand to 100% the use of `new_index` which we are returning. Could you please elaborate more on that?

jannikluhn:
If you deserialize a composite structure (a list or a container), you don't always know the length of the elements before actually deserializing them individually (e.g. a list of bytes, all may have a different size). But the overlying list deserializer needs to know at which location in the string the elements start. That's why the `new_index` is returned.

In RLP you don't need to know the types to do this, there you can infer the lengths of the elements just from the structure of the input. This means that all of this is done in the encoding/decoding stage, so the `sedes` objects don't have to deal with this.

I'm not sure how this affects `py-ssz`. Maybe the distinction between encoding/decoding and serializing/deserializing is not that good an idea here. Maybe it still makes sense and just has to be tweaked slightly. I guess you just have to play around with the design a little and see what feels right.

[...]

PR #7, opened 2018-12-05
Add serialization/deserialization functionality for Integers

Bhargavasomu:
## What was wrong?
As part of Issue #4 , `Integer` types are now serializable and deserializable. This includes both `int<N>` and `uint<N>`.

## How was it fixed?
A new class was created for the Integers which support both `int<N>` and `uint<N>`. This class was infered by the `infer_sedes` method.

#### Cute Animal Picture


Bhargavasomu:
@jannikluhn @hwwhww could you please take a look. It is working when tested in `ipython`. I will be adding the `tests` as part of the PR in a while.

Bhargavasomu:
Added the tests too now. Please let me know if any other tests need to be added.

ssz/codec.py
```diff
@@ -10,11 +10,13 @@
 )
 
 
-def encode(obj):
+def encode(obj, serializer_type=None):
```
jannikluhn: This should not default to `None` anymore as it's required now.

To be consistent with the terminology, calling it `sedes` would be better I think.

Bhargavasomu:
Ok then, I'm gonna make the above changes as well as remove the support for `signed integers`

ssz/sedes/integer.py
```diff
@@ -0,0 +1,67 @@
...
+)
+
+
+class Integer:
+    """
+    A sedes for integers (Handles both int<N> and uint<N>).
+    """
+    # This variable captures 'N' in 'int<N>' or 'uint<N>'
+    num_bytes = 0
+    allows_signed_ints = True
```
hwwhww: 1. Suggestion: rename to `is_signed`
2. Set to unsigned by default.

ssz/utils.py
```diff
@@ -12,12 +13,26 @@ def is_sedes(obj):
...
 
 
-def infer_sedes(obj):
+def get_integer_serializer(serializer_type):
+    '''
+    Given serializer_type as 'int<N>' or 'uint<N>', this function
+    returns the corresponding serializer object.
+    '''
+    integer_serializer = Integer(serializer_type)
+    return integer_serializer
```
hwwhww: Maybe just 
```suggestion
    return Integer(serializer_type)
```

ssz/utils.py
```diff
@@ -12,12 +13,26 @@ def is_sedes(obj):
...
     """
     Try to find a sedes objects suitable for a given Python object.
+    :param serializer_type: str
+    The serializer_type parameter is optional and can be only
+    of the format 'int<N>' or 'uint<N>'(as of now)
     """
     if isinstance(obj, bool):
         return boolean
+    elif isinstance(obj, int):
+        return get_integer_serializer(serializer_type)
```
hwwhww: Super *slightly* optimization for eth2.0 use case: maybe swap the order of `isinstance(obj, int)` and `isinstance(obj, bool)`.
```python
    if isinstance(obj, int):
        return get_integer_serializer(serializer_type)
    elif isinstance(obj, bool):
        return boolean
```

ssz/sedes/integer.py
Bhargavasomu: @hwwhww I have removed the support for `signed integers`, as suggested by jannikluhn. So this won't be a matter anyway. But I could put back the support for signed integers if you want. What's your call?

ssz/utils.py
Bhargavasomu: @hwwhww but if the code was refactored as above, the following bug would occur where `bool` type would be regarded as `int`. Because as per `python3.6`
```
isinstance(True, int) = True
```

Bhargavasomu:
@jannikluhn  @hwwhww I have made the changes specified, could you please review it again. Thankyou.

ssz/sedes/integer.py
hwwhww: I think it's okay to remove signed integers. 👍

hwwhww:
It occurred to me that we should add type hinting. 😮

Bhargavasomu:
I would be glad to do the type hinting, but maybe part of another issue which I would be taking up. This is because while building the module, I want to primarily focus on the code design and structure.

ssz/sedes/integer.py
```diff
@@ -6,62 +6,51 @@
...
+        try:
+            serialized_obj = val.to_bytes(self.num_bytes, 'big')
+        except OverflowError as err:
+            raise SerializationError('As per specified sedes object, %s' % err, val)
 
-        return val.to_bytes(self.num_bytes, 'big', signed=self.allows_signed_ints)
+        return serialized_obj
 
-    def _deserialize(self, data, start_idx):
+    def deserialize_from(self, data, start_index):
```
jannikluhn: Sorry for changing my mind here, but I don't like the `from` now that I see it. It sounds like it refers to `data` instead of the `start_index`. `slice` would be more clear, but it clashes with the Python object. What do you think of `deserialize_segment`?

ssz/codec.py
```diff
@@ -4,19 +4,37 @@
...
-    `serializer_type` needs to be explicitly mentioned for encode/decode
-    of integers(as of now)
+    `sedes` needs to be explicitly mentioned for encode/decode
+    of integers(as of now).
+    `sedes` parameter could be given as a string or as the
+    actual sedes object itself.
     """
-    serialized_obj = infer_sedes(obj, serializer_type).serialize(obj)
+    if sedes:
+        if isinstance(sedes, str):
```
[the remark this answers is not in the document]
Bhargavasomu: @jannikluhn but this approach wouldn't be generic enough for the new `uintN` type, for which we didn't hard code. For example, what if a user wants to give `uint24`(rare but possible), but we wouldn't hardcode that, which would lead to a `InvalidSedesError`. So how do we handle these use cases?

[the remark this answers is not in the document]
jannikluhn: We can hardcode all common values (e.g. everything between 8 and 64 as well as the values used in the spec).

Bhargavasomu:
@jannikluhn I have made the necessary changes. Could you please take a look now? Thanks.

Bhargavasomu:
@hwwhww @jannikluhn assuming that the part of `unsigned integers` was ok, I went on to create the support  for `serialization/deserialization` of `hashN` and `address`. Please take a look and let me know for any changes required. Thanks

Bhargavasomu:
@jannikluhn @hwwhww made the changes, please check it now. Thanks!

PR #10, opened 2018-12-11
List

Bhargavasomu:
## What was wrong?
Implementing `serialization/deserialization` for `lists`
Issue #4 

## How was it fixed?
By following the implementation details in the spec

#### Cute Animal Picture

### DIFF
--- a/ssz/sedes/boolean.py
+++ b/ssz/sedes/boolean.py
@@ -19,16 +19,40 @@ class Boolean:
         else:
             raise Exception("Invariant: no other options for boolean values")
 
-    def deserialize(self, serialized_obj):
-        if serialized_obj == b'\x00':
-            return False
-        elif serialized_obj == b'\x01':
-            return True
+    def deserialize_segment(self, serialized_obj, start_index):
+        """
+        Deserialize the data from the given start_index
+        """
+        # Make sure we have sufficient data for inferring length of list
+        if len(serialized_obj) < start_index + 1:
+            raise DeserializationError(
+                'Insufficient data for deserializing',
+                serialized_obj
+            )
+
+        # Deal with only the first byte of the whole serialized data
+        boolean_serialized_obj = serialized_obj[start_index: start_index + 1]
+        if boolean_serialized_obj == b'\x00':
+            deserialized_obj = False
+        elif boolean_serialized_obj == b'\x01':
+            deserialized_obj = True
         else:
             raise DeserializationError(
                 'Invalid serialized boolean.  Must be either 0x01 or 0x00',
                 serialized_obj
             )
+
+        return deserialized_obj, start_index + 1
+
+    def deserialize(self, data):
+        deserialized_data, end_index = self.deserialize_segment(data, 0)
+        if end_index != len(data):
+            raise DeserializationError(
+                'Data to be deserialized is too long',
+                data
+            )
+
+        return deserialized_data
 
 
 boolean = Boolean()
--- a/ssz/sedes/list.py
+++ b/ssz/sedes/list.py
@@ -1,3 +1,8 @@
+from collections.abc import (
+    Iterable,
+    Mapping,
+)
+
 from ssz.exceptions import (
     DeserializationError,
     SerializationError,
@@ -16,28 +21,42 @@ class List:
     """
     LENGTH_BYTES = 4
 
-    def __init__(self, sedes):
-        # This sedes object corresponds to each item of the list
-        self.sedes = sedes
+    def __init__(self, element_sedes):
+        # This sedes object corresponds to each element of the iterable
+        self.element_sedes = element_sedes
 
     def serialize(self, val):
-        if not isinstance(val, list):
-            raise SerializationError('Can only serialize lists', val)
-        # Make sure all items in list are of same type
-        if not all((type(item)) == (type(val[0])) for item in val):
-            raise SerializationError('Can only serialize lists having elements of same type', val)
+        if not isinstance(val, Iterable) or isinstance(val, Mapping):
+            raise SerializationError(
+                'Can only serialize Iterable objects, except Dictionaries',
+                val
+            )
 
-        serialized_list_string = b''
+        serialized_iterable_string = b''
+        element_type = None
         for item in val:
-            serialized_list_string += self.sedes.serialize(item)
-        if len(serialized_list_string) >= 2 ** (self.LENGTH_BYTES * 8):
+            # Make sure the items in the iterable are of same type
+            if element_type is None:
+                # First update the element_type if None
+                element_type = type(item)
+            else:
+                if type(item) != element_type:
+                    raise SerializationError(
+                        'Can only serialize lists having elements of same type',
+                        val
+                    )
+
+            # Serialization after checking the consistency of list item
+            serialized_iterable_string += self.element_sedes.serialize(item)
+
+        if len(serialized_iterable_string) >= 2 ** (self.LENGTH_BYTES * 8):
             raise SerializationError(
                 'List too long to fit into {} bytes after serialization'.format(self.LENGTH_BYTES),
                 val
             )
-        serialized_len = len(serialized_list_string).to_bytes(self.LENGTH_BYTES, 'big')
+        serialized_len = len(serialized_iterable_string).to_bytes(self.LENGTH_BYTES, 'big')
 
-        return serialized_len + serialized_list_string
+        return serialized_len + serialized_iterable_string
 
     def deserialize_segment(self, data, start_index):
         """
@@ -61,10 +80,11 @@ class List:
             )
 
         deserialized_list = []
-        item_index = start_index + self.LENGTH_BYTES
-        while item_index < list_end_index:
-            object, item_index = self.sedes.deserialize_segment(data, item_index)
-            deserialized_list.append(object)
+        # element_start_index is the start index of an element in the serialized bytes string
+        element_start_index = start_index + self.LENGTH_BYTES
+        while element_start_index < list_end_index:
+            element, element_start_index = self.element_sedes.deserialize_segment(data, element_start_index)  # noqa: E501
+            deserialized_list.append(element)
 
         return deserialized_list, list_end_index
 
@@ -79,7 +99,7 @@ class List:
         return deserialized_data
 
 
-addr_list = List(address)
-bool_list = List(boolean)
-hash_list = List(hash32)
-int_list = List(uint32)
+address_list = List(address)
+boolean_list = List(boolean)
+hash32_list = List(hash32)
+uint32_list = List(uint32)
--- a/ssz/sedes/list.py
+++ b/ssz/sedes/list.py
@@ -1,6 +1,5 @@
 from collections.abc import (
     Iterable,
-    Mapping,
 )
 
 from ssz.exceptions import (
@@ -29,28 +28,15 @@ class List:
         self.element_sedes = element_sedes
 
     def serialize(self, val):
-        if not isinstance(val, Iterable) or isinstance(val, Mapping):
+        if not isinstance(val, Iterable):
             raise SerializationError(
                 'Can only serialize Iterable objects, except Dictionaries',
                 val
             )
 
-        serialized_iterable_string = b''
-        element_type = None
-        for item in val:
-            # Make sure the items in the iterable are of same type
-            if element_type is None:
-                # First update the element_type if None
-                element_type = type(item)
-            else:
-                if type(item) != element_type:
-                    raise SerializationError(
-                        'Can only serialize lists having elements of same type',
-                        val
-                    )
-
-            # Serialization after checking the consistency of list item
-            serialized_iterable_string += self.element_sedes.serialize(item)
+        serialized_iterable_string = b"".join(
+            self.element_sedes.serialize(element) for element in val
+        )
 
         if len(serialized_iterable_string) >= 2 ** (self.LENGTH_BYTES * 8):
             raise SerializationError(
@@ -86,7 +72,9 @@ class List:
         # element_start_index is the start index of an element in the serialized bytes string
         element_start_index = start_index + self.LENGTH_BYTES
         while element_start_index < list_end_index:
-            element, element_start_index = self.element_sedes.deserialize_segment(data, element_start_index)  # noqa: E501
+            element, element_start_index = self.element_sedes.deserialize_segment(
+                data, element_start_index
+            )
             deserialized_list.append(element)
 
         return deserialized_list, list_end_index
--- a/ssz/sedes/list.py
+++ b/ssz/sedes/list.py
@@ -28,7 +28,12 @@ class List:
         self.element_sedes = element_sedes
 
     def serialize(self, val):
-        if not isinstance(val, Iterable):
+        if (
+                not isinstance(val, Iterable) or
+                isinstance(val, bytes) or
+                isinstance(val, bytes) or
+                isinstance(val, str)
+        ):
             raise SerializationError(
                 'Can only serialize Iterable objects, except Dictionaries',
                 val
@@ -92,5 +97,6 @@ class List:
 
 address_list = List(address)
 boolean_list = List(boolean)
+empty_list = List(None)
 hash32_list = List(hash32)
 uint32_list = List(uint32)
--- a/ssz/sedes/list.py
+++ b/ssz/sedes/list.py
@@ -23,9 +23,21 @@ class List:
     """
     LENGTH_BYTES = 4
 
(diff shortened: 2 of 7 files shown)


Bhargavasomu:
@jannikluhn I have still not added the tests yet. But as far as the design is concerned, could you please take a look. Also one thing to note is that, we haven't added the function `deserialize_segment` for the `boolean sedes`. This is needed in the case of `boolean lists`. Will make that change and later add the `tests`.

ssz/sedes/list.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, sedes):
+        # This sedes object corresponds to each item of the list
+        self.sedes = sedes
+
+    def serialize(self, val):
+        if not isinstance(val, list):
+            raise SerializationError('Can only serialize lists', val)
+        # Make sure all items in list are of same type
+        if not all((type(item)) == (type(val[0])) for item in val):
```
jannikluhn: This must not be checked here, but in the loop where serialization happens. Also, indexing doesn't work for iterables in general, and even a list might be empty.

ssz/sedes/list.py
```diff
@@ -0,0 +1,85 @@
...
+        list_end_index = start_index + self.LENGTH_BYTES + list_length
+        # Make sure we have sufficent data for inferring the whole list
+        if len(data) < list_end_index:
+            raise DeserializationError(
+                'Insufficient data: Cannot retrieve the whole list data',
+                data
+            )
+
+        deserialized_list = []
+        item_index = start_index + self.LENGTH_BYTES
```
jannikluhn: `item_index` makes it sound like it's an index in the serialized list, not `data`. Just `data_index` would be better, but sounds weird. Other suggestions?

ssz/sedes/list.py
```diff
@@ -0,0 +1,85 @@
...
+        while item_index < list_end_index:
+            object, item_index = self.sedes.deserialize_segment(data, item_index)
```
jannikluhn:
```suggestion
            element, item_index = self.sedes.deserialize_segment(data, item_index)
```
as `object` is a built in.

ssz/utils.py
```diff
@@ -26,6 +26,11 @@ def infer_sedes(obj):
             'uint sedes object or uint string needs to be specified for ints',
             obj
         )
+    elif isinstance(obj, list):
```
```suggestion
    elif isinstance(obj, Iterable):
```

Bhargavasomu:
@jannikluhn I also like the idea of replacing `list` with `Iterable`. This would add support for `lists`, `tuples` and `sets` in a single go. But the same functionality would also inherently add support for `dictionaries`. But as far as the specs go, we have a different logic for serializing/deserializing `dictionaries`. So should we put a check for this special case of `dictionaries`?

jannikluhn:
Good point. I think we should distinguish two situations: Serialization and sedes inference. Serialization

- Serialization: I don't see a reason why we would want to restrict that. If one explicitly asks for serializing a dictionary as a list and provides a suitable `element_sedes` it should work.

- Sedes inference: This wouldn't work for containers anyways, and for lists it would fail when trying to infer the element sedes. But it might be a good idea to catch `Mapping`s early in order to get better error messages. But this isn't implemented yet anyways.

ssz/sedes/list.py
Bhargavasomu: @jannikluhn how about `element_start_index`?

jannikluhn: I like `element_start_index` :+1:

Bhargavasomu:
@jannikluhn fixed them all. Will now add `tests`. Please review the changes in the meanwhile.

Bhargavasomu:
@jannikluhn one more thing with `sets` type is that the order of iteration changes from time to time. So I don't think it is consistent to add support for that. What do you say?

jannikluhn:
I agree that this is suboptimal, but AFAIK there's no general way to check for orderliness and treating `set`s as a special case doesn't feel right. Also, serializing `set`s is not inherently wrong. So I think just adding a warning to the docstring of the list sedes object is sufficient.

Bhargavasomu:
@jannikluhn can we remove `flake8` on `tests`? I don't think it is needed.

Bhargavasomu:
@jannikluhn but apart from the linting, I have added the testcases too. Please have a look

ssz/utils.py
```diff
@@ -26,6 +30,11 @@ def infer_sedes(obj):
             'uint sedes object or uint string needs to be specified for ints',
             obj
         )
+    elif isinstance(obj, Iterable):
+        raise SerializationError(
```
jannikluhn: Noticing this just now: `SerializationError` is probably not the right choice here. `pyrlp` uses `TypeError`.

tests/core/test_list_serializer.py
```diff
@@ -0,0 +1,218 @@
...
+        ((), uint32_list, b'\x00\x00\x00\x00'),
+        ([], hash32_list, b'\x00\x00\x00\x00'),
+        ((), hash32_list, b'\x00\x00\x00\x00'),
+        ([], address_list, b'\x00\x00\x00\x00'),
+        ((), address_list, b'\x00\x00\x00\x00'),
+        ([], boolean_list, b'\x00\x00\x00\x00'),
+        ((), boolean_list, b'\x00\x00\x00\x00'),
+
+        # Serialize uint32 Iterables
+        ([0, 1, 2, 3, 4], uint32_list, uint32.serialize(20) + uint32.serialize(0) + uint32.serialize(1) + uint32.serialize(2) + uint32.serialize(3) + uint32.serialize(4)),
```
jannikluhn: I think we should put the actual bytes here, not construct it dynamically.

jannikluhn:
> can we remove flake8 on tests? I don't think it is needed as the tests while testing with hash32 etc would be big.

No, flake8 is useful for tests too. I'd be fine with disabling it for individual lines though if there's a good reason why it's so long and breaking it up would be worse. But this isn't the case here I think. For instance, for long byte strings we could use the following pattern:

```Py
(
    b"\x00\x00\x00\x40"
    b"\x00" * 32,
    b"\x11" * 32,
)

File: ssz/sedes/boolean.py [hunk 7]
```diff
@@ -19,16 +19,40 @@ def serialize(self, obj):
...
+        """
+        # Make sure we have sufficient data for inferring length of list
+        if len(serialized_obj) < start_index + 1:
+            raise DeserializationError(
+                'Insufficient data for deserializing',
+                serialized_obj
+            )
+
+        # Deal with only the first byte of the whole serialized data
+        boolean_serialized_obj = serialized_obj[start_index: start_index + 1]
```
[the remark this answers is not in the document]
Bhargavasomu: @jannikluhn this won't work because of the following example.
Say the string is `str = \x00\x01`. Now if we want to get the first byte, then
```
str[0] = 0
str[0:1] = \x00
```
And the `boolean sedes` recognizes only `\x00` or `\x01` for deserialization.

Bhargavasomu:
@jannikluhn all done, but there seems to be some `setup` issue with `circle-ci` regarding `pytest`. Didn't understand what it is though. And don't remember changing any setup for this to occur. Could you please look into it?

hwwhww:
@Bhargavasomu fixed!

hwwhww:
@jannikluhn @Bhargavasomu @pipermerriam 

If we follow the convention of Py-EVM, should the deserialized result be `tuple` by default?

ssz/utils.py
```diff
@@ -2,11 +2,14 @@
     Iterable,
 )
 
-from ssz.exceptions import (
-    SerializationError,
+from toolz import (
```
jannikluhn: ```from eth_utils.toolz import (```

(following a py-evm convention that I just learned about)

jannikluhn:
> If we follow the convention of Py-EVM, should the deserialized result be tuple by default?

No preference for either from my side. `list` sounds natural as that's how they're called in the SSZ spec. But if it tuples would be a better fit for py-evm I'm fine with this as well.

Bhargavasomu:
@jannikluhn done

ssz/sedes/list.py
```diff
@@ -0,0 +1,96 @@
...
+    (This is because iteration in sets doesn't always happen in the same order)
+    """
+    LENGTH_BYTES = 4
+
+    def __init__(self, element_sedes):
+        # This sedes object corresponds to each element of the iterable
+        self.element_sedes = element_sedes
+
+    def serialize(self, val):
+        if not isinstance(val, Iterable):
```
pipermerriam: If you change this to use `collections.abc.Sequence` then it will disallow non-ordered iterable types like `set` or `dict`.  In addition, you should potentially disallow all of `bytes,bytearray,str` since they will be *interpreted* as an iterable of their individual characters which is almost definitely not the desired/expected behavior.

ssz/utils.py
```diff
@@ -15,17 +22,38 @@ def is_sedes(obj):
     return hasattr(obj, 'serialize') and hasattr(obj, 'deserialize')
 
 
+def infer_list_sedes(obj):
+    try:
+        first_element, iterator = peek(obj)
+    except StopIteration:
+        # For empty lists we can use any element sedes.
+        # Using uint32_list here as the default
```
pipermerriam: Since this is arbitrary, what do you think about having a special case sedes which *cannot* actually be used to prevent accidental mistakes/weirdness.  Something like `EmptyListSedes` which just always returns an empty list....

ssz/utils.py
```diff
@@ -15,17 +22,38 @@ def is_sedes(obj):
...
     """
     if isinstance(obj, bool):
         return boolean
     elif isinstance(obj, int):
-        raise SerializationError(
+        raise TypeError(
             'uint sedes object or uint string needs to be specified for ints',
             obj
         )
+    elif isinstance(obj, Iterable):
```
pipermerriam: another place for `Sequence` over `Iterable`

ssz/sedes/list.py
jannikluhn: Agree with the string like checks. But I'd rather keep the `Iterable` as `Sequence` would exclude generators which might be nice to have in some cases.

Bhargavasomu:
@jannikluhn I have added `empty_list` sedes object. Please have a look. I didn't fully understand about the `tuple` default thing. Is it that we should replace the whole `list` sedes thing with the `tuple` sedes object which also supports serializing/deserializing `lists`?

pipermerriam:
@jannikluhn regarding the idea of using `tuples` as the default instead of `list`.  Maybe the `List` sedes should be renamed to `Array` to remove any misnomers on what type it corresponds to.

@Bhargavasomu the `List` sedes is really just intended to encode "arrays", or "ordered collections" of values.  Currently they return a python `list` type.  I'm only advocating that their return type be changed to `tuple` in favor of immutability.

Bhargavasomu:
Is there any way to explicitly find if an object corresponds to a `Generator`? I did try using `isinstance(obj, types.GeneratorType)`, but it wan't detecting `range(10)` as a generator function.

jannikluhn:
That's because `range` is not a generator. It's _probably_ a bad idea to check this anyway, why do you want to do this?

Bhargavasomu:
@jannikluhn I wanted to explicitly check for `Generators` becuase it's only for those generators, that we want to support all `Iterables`, but I'm inclined on supporting only `Sequences` and `Generators`, as per what @pipermerriam had mentioned. What do you say?

I understand `range` is not a generator, and we may need to changing the test cases to include `(i for i in range(5))`. Would that be ok?

jannikluhn:
I would rather not do this as we will probably miss some `Iterable`s that would be nice to support. I'd be fine with explicitly disallowing `set`s and (unordered) `dict`s if we want to be safe (even though I don't think that's strictly necessary).

Bhargavasomu:
@jannikluhn ok then, is there anymore changes I need to do, before this gets merged?

ssz/sedes/list.py
```diff
@@ -0,0 +1,102 @@
...
+class List:
+    """
+    A sedes for lists.
+
+    WARNING: Avoid sets if possible, may not always lead to expected results
+    (This is because iteration in sets doesn't always happen in the same order)
+    """
+    LENGTH_BYTES = 4
+
+    def __init__(self, element_sedes):
```
jannikluhn: If I understand correctly, the empty list sedes just uses `None` as `element_sedes` and trying to serialize/deserialize something non-empty would raise an `AttributeError` which is misleading.

I see two better options, maybe there are more:

1) Add an `EmptyList` sedes that is not related to `List` at all and only works for empty lists (raising `Serialization`/`DeserializationError`s for everything else).
2) Add an explicit argument `empty` to `List.__init__` and check that during serialization/deserialization. Also raise `ValueError`s if one tries to initialize `List` with `empty` and an `element_sedes` or without `empty` and without an `element_sedes`.

ssz/sedes/list.py
```diff
@@ -0,0 +1,102 @@
...
+        deserialized_list = []
+        # element_start_index is the start index of an element in the serialized bytes string
+        element_start_index = start_index + self.LENGTH_BYTES
+        while element_start_index < list_end_index:
+            element, element_start_index = self.element_sedes.deserialize_segment(
+                data, element_start_index
+            )
+            deserialized_list.append(element)
+
+        return deserialized_list, list_end_index
```
jannikluhn:
```suggestion
        return tuple(deserialized_list), list_end_index
```

Bhargavasomu:
@jannikluhn done! Please review once again. Thankyou

pipermerriam:
I believe this type check works and does what we want.

```python
from collections.abc import Generator, Sequence

is_allowed_iterable(value):
   return isinstance(value, (Generator, Sequence))
```

ssz/sedes/list.py
```diff
@@ -0,0 +1,119 @@
...
+        # This sedes object corresponds to each element of the iterable
+        self.element_sedes = element_sedes
+        # This empty bool indicates whether this sedes is meant for empty lists
+        self.empty = empty
+
+    def serialize(self, val):
+        if (
+                not isinstance(val, Iterable) or
+                isinstance(val, bytes) or
+                isinstance(val, bytes) or
```
pipermerriam: This check is repeated twice.

ssz/sedes/list.py
```diff
@@ -0,0 +1,119 @@
...
+                isinstance(val, str)
```
pipermerriam: Easy to combine this as `isinstance(val, (str, bytes))`

ssz/sedes/list.py
```diff
@@ -0,0 +1,119 @@
...
+        if len(data) < list_end_index:
+            raise DeserializationError(
+                'Insufficient data: Cannot retrieve the whole list data',
+                data
+            )
+
+        deserialized_list = []
+        # element_start_index is the start index of an element in the serialized bytes string
+        element_start_index = start_index + self.LENGTH_BYTES
+        while element_start_index < list_end_index:
```
pipermerriam: This `while` loop makes me *mildly* uncomfortable.  It'd be really nice to figure out a way to have stronger guarantees about the exit conditions.  One that I can think of would be to enforce that each subsequent `element_start_index` **must** be greater than the previous one, so that we have a strong guarantee that this will eventually hit the end of the `data`.

ssz/sedes/list.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+        return deserialized_data
+
+
+address_list = List(address)
+boolean_list = List(boolean)
+empty_list = List(empty=True)
+hash32_list = List(hash32)
+uint32_list = List(uint32)
```
pipermerriam: I think these should probably be defined in `ssz/sedes/__init__.py` so that we don't have to import `address, boolean, hash32, int32` in this module.

ssz/utils.py
```diff
@@ -15,17 +22,37 @@ def is_sedes(obj):
     return hasattr(obj, 'serialize') and hasattr(obj, 'deserialize')
 
 
+def infer_list_sedes(obj):
+    try:
+        first_element, iterator = peek(obj)
```
pipermerriam: since `iterator` isn't used I'd recommend renaming this variable to `_`

ssz/sedes/list.py
Bhargavasomu: @pipermerriam this was noticed and has been fixed as part of this PR https://github.com/ethereum/py-ssz/pull/13

pipermerriam:
@jannikluhn no opposition from me on that.  Good points.  I'd *highly* recommend adding a test for every API that can accept a generator to test the generator use case to be sure that code doesn't sneak in which tries to consume the input iterable multiple times.

PR #13, opened 2018-12-17
Add support for Bytes and Bytearray objects

Bhargavasomu:
## What was wrong?
Bytes Objects are now supported as part of Issue #4 

#### Cute Animal Picture


Bhargavasomu:
@jannikluhn could you please review this. I am very keen on not hardcoding the value `4` in the `bytes` sedes part. But in the specs the variable used is `LENGTH_BYTES` which I feel could be misleading here. Any suggestions on that?
    Also I will be adding support to `byte_lists` and also be adding test cases to that in a moment.

Bhargavasomu:
@jannikluhn moved the `constants` to a seperate file now. Please take a look. Thankyou!

hwwhww:
> I am very keen on not hardcoding the value `4` in the `bytes` sedes part. But in the specs the variable used is `LENGTH_BYTES` which I feel could be misleading here.

Agreed +1
And I just noticed that the description on spec: `byte_length = (len(value)).to_bytes(LENGTH_BYTES, 'big')` is hard to understand... 😂

@jannikluhn @Bhargavasomu 
Would either of you like to raise an issue on eth2.0-specs?

ssz/utils.py
```diff
@@ -46,11 +47,16 @@ def infer_sedes(obj):
...
     if isinstance(obj, bool):
         return boolean
+
     elif isinstance(obj, int):
         raise TypeError(
             'uint sedes object or uint string needs to be specified for ints',
             obj
         )
+
+    elif isinstance(obj, bytes) or isinstance(obj, bytearray):
```
```suggestion
    elif isinstance(obj, (bytes, bytearray)):
```

ssz/sedes/bytes.py
```diff
@@ -0,0 +1,72 @@
...
+            raise SerializationError(
+                "Can only serialize bytes or bytearray objects",
+                val
+            )
+
+        object_len = len(val)
+        if object_len >= 2 ** (BYTES_PREFIX_LENGTH * 8):
+            raise SerializationError(
+                'Object too long for its length to fit into {} bytes after serialization'
+                .format(BYTES_PREFIX_LENGTH),
```
jannikluhn: Not sure about this line break. Consider using f-strings like this:
```Py
            raise SerializationError(
                f'Object too long for its length to fit into {BYTES_PREFIX_LENGTH} bytes "
                f"after serialization',
                val,
            )
```

Bhargavasomu:
@jannikluhn done now. Please check it once again. Thankyou.

Issue #9, opened 2018-12-07
Introduce abstract base class for sedes objects


jannikluhn:
We currently check if an object is a `sedes` by checking if it has both a `serialize` and a `deserialize` method: https://github.com/ethereum/py-ssz/blob/master/ssz/utils.py#L12

This is ugly and can in theory lead to bugs if a wrong object with the same interface is used (e.g. an RLP sedes).

As [suggested](https://github.com/ethereum/py-ssz/pull/3#discussion_r238741262) by @pipermerriam, we could instead use an abstract base class and replace the `hasattr` checks with `isinstance` checks against it. Then we could also get rid of the `is_sedes` function altogether as `isinstance(sedes, BaseSedes)` would be equally clear.

ChihChengLiang:
work on this

ChihChengLiang:
Question: Do we need to add a new abstract class for list_sedes, say List and Bytes?

PR #16, opened 2018-12-19
Fix #15, add tree hash

### LINKED ISSUE #15: Implement tree hashing
Implement tree hashing as defined [here](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash).

Suggested API: `tree_hash(obj, sedes=None)`. If no sedes object is given, it should be inferred using `infer_sedes`. By checking which type of sedes object is given or inferred (using `isinstance`), the call can be delegated to the appropriate implementation.

Note that containers are not implemented yet, so for now only base types and lists will work.

ChihChengLiang:
## What was wrong?

Issue #15 

## How was it fixed?

- Add a tree hash module
- Add `hash_eth2`
- Add `merkle_hash`

#### Cute Animal Picture


ssz/tree_hash/merkle_hash.py
```diff
@@ -0,0 +1,41 @@
...
+from eth_typing import (
+    Hash32,
+)
+
+from .hash_eth2 import (
+    hash_eth2,
+)
+
+
+def merkle_hash(input_items: Sequence[Any], chunk_size: int=128) -> Hash32:
```
hwwhww: I think we should add `SSZ_CHUNK_SIZE = 128` in `ssz/constants.py`.

ssz/tree_hash/merkle_hash.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Merkle tree hash of a list of homogenous, non-empty items
+    """
+
+    data_length = len(input_items).to_bytes(32, 'big')
```
hwwhww: (for the whole function) maybe bring back the comments from the spec?

ssz/tree_hash/tree_hash.py
```diff
@@ -0,0 +1,45 @@
...
+        sedes = infer_sedes(input_object)
+
+    if isinstance(sedes, (Boolean, Hash, UnsignedInteger)):
+        serialization = sedes.serialize(input_object)
+        if isinstance(sedes, UnsignedInteger) and sedes.num_bytes > 32:
+            return hash_eth2(serialization)
+        else:
+            return serialization.ljust(32, b'\x00')
+
+    if isinstance(input_object, list):
```
hwwhww: I think it's not only `list`. We want to support both `list` and `tuple`. Maybe:
```python
from collections import abc

if isinstance(input_object, abc.Iterable):
    return merkle_hash([hash_tree_root(item, sedes.element_sedes) for item in input_object])

```

tests/tree_hash/test_tree_hash.py
```diff
@@ -0,0 +1,41 @@
...
+    'value,sedes',
+    (
+        (0, uint16),
+        (56, uint16),
+        ([123, 456, 789], uint32_list),
+        (55665566, uint512),
+        ([], uint32_list),
+        ([b'\x56' * 32], hash32_list),
+        ([b'\x56' * 32] * 100, hash32_list),
+        ([b'\x56' * 32] * 101, hash32_list),
```
hwwhww: Also, test the tuple inputs.

ssz/tree_hash/tree_hash.py
```diff
@@ -0,0 +1,45 @@
...
+
+from .hash_eth2 import (
+    hash_eth2,
+)
+from .merkle_hash import (
+    merkle_hash,
+)
+
+
+def hash_tree_root(input_object: Any, sedes: Any=None) -> Hash32:
```
hwwhww: Note: with #9, we will replace `sedes: Any` with `sedes: BaseSedes`.

Issue #14, opened 2018-12-18
Implement containers


jannikluhn:
Implement serialization and deserialization of SSZ containers as specified [here](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#container).

The API should be similar to `pyrlp`. In particular, there should be a class `ssz.Serializable` that can be subclassed to define SSZ containers. Subclasses have a `fields` attribute that defines their internal structure (see [this class](https://github.com/ethereum/py-evm/blob/master/eth/beacon/types/attestations.py#L21) for an example). `ssz.Serializable` classes are sedes objects for their instances, so they should define `serialize` and `deserialize` classmethods. Field values are initialized using the `__init__` constructor that checks that all, but no additional arguments are given. Field values can be accessed as read only properties. The code should be well tested.

I suspect that pretty much all of this except for the implementation of `serialize` and `deserialize` can be copied from `pyrlp`. The `Changeset` functionality is optional.

Bhargavasomu:
On this.

Bhargavasomu:
@jannikluhn I have observed that the `serializable` class in `pyrlp` has a lot of functions like `__hash__`, and `__iter__`. I don't gues such functions are required here right?

Bhargavasomu:
@hwwhww @jannikluhn I have gone through the specs of `container` and while doing the implementation, I have observed that there is no need for `sorting` the `fields`. Please confirm this.

hwwhww:
> I have observed that there is no need for sorting the fields. 

Can you elaborate more?
Do you mean any incompatibility to [the spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#container) - "To serialize a container, obtain the set of its field's names and sort them lexicographically"?

Bhargavasomu:
@hwwhww yes, the same `serialization/deserialization` process would work the same with and without sorting, as the `metaclass` stores them in the order given by the user (need not be sorted order). And finally the same sorted order can be used for `deserialization` also. So in `contrast` to the `spec`, I don't see the use and need for `sorting`.

That said, please correct me if I am missing something.

jannikluhn:
> @jannikluhn I have observed that the serializable class in pyrlp has a lot of functions like __hash__, and __iter__. I don't gues such functions are required here right?

[...]

PR #17, opened 2018-12-20
Support for Serialization/Deserialization of Containers

Bhargavasomu:
## What was wrong?
Fixes Issue #14 

## How was it fixed?
Overall structure of the code for container was copied from `pyrlp`. The `serialization` and `deserialization` functionalities have been changed so as to suit the `specs` of `ssz`. Test cases have also been copied and adjusted from `pyrlp`. I have added a few more though. Since the `spec` has been updated, the fields have not been sorted.

#### Cute Animal Picture


Bhargavasomu:
@hwwhww @jannikluhn please review this.

tests/core/test_serializable.py
```diff
@@ -0,0 +1,484 @@
...
+    bytes_sedes,
+    uint32,
+    uint32_list,
+)
+from ssz.utils import (
+    infer_sedes,
+)
+
+
+class RLPType1(Serializable):
```
hwwhww: All those class names should be `SSZ*`.

tests/core/test_serializable.py
```diff
@@ -0,0 +1,484 @@
...
+    return _type_2.copy()
+
+
+@pytest.fixture(params=[_type_1_a, _type_1_b, _type_2])
+def ssz_obj(request):
+    return request.param.copy()
+
+
+@pytest.mark.parametrize(
+    'rlptype,args,kwargs,exception_includes',
```
hwwhww: Not RLP.

tests/core/test_serializable.py
```diff
@@ -0,0 +1,484 @@
...
+        (b'\x00\x00\x00\x08\x00\x00\x00\x00\x00\x00\x00\x01' + b'\x00', RLPType3),
+        (b'\x00\x00\x00\x08\x00\x00\x00\x00\x00\x00\x00\x01' + b'\x00', RLPType4),
+    ),
+)
+def test_container_deserialize_bad_values(value, sedes):
+    with pytest.raises(DeserializationError):
+        sedes.deserialize(value)
+
+
+def test_serializable_serialization(type_1_a, type_1_b, type_2):
```
hwwhww: 1. I think you can use `@pytest.mark.parametrize`?
2. Where are these result came from?

Bhargavasomu: @hwwhww yes, I will parametrize them. The results were generated using `SSZ` and not `RLP`. I took extra care there.

Bhargavasomu:
@hwwhww done now, please check again. Thankyou.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,420 @@
...
+            raise DeserializationError(
+                'Insufficient data: Cannot retrieve the whole container',
+                data
+            )
+
+        deserialized_field_values = []
+        field_start_index = start_index + CONTAINER_PREFIX_LENGTH
+        # sedes_generator generates the sedes object for the fields in order
+        sedes_generator = (field_sedes for field_name, field_sedes in cls._meta.fields)
+        while field_start_index < container_end_index:
```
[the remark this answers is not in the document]
jannikluhn: You're right. But what can still happen is that there's some junk bytes between the end of the last element and the end of the container (as we don't check that the end index of the last element is on the end index of the container). That's independent of the loop though.

I still think that iterating over the fields would make the code easier to understand though. After all, what we're doing is deserializing the elements one after another; we're not deserializing until we have no bytes left (at least in my mind). That would also be analogous to the serialization code.

Bhargavasomu:
@jannikluhn done now, please check again. Thankyou.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+def _get_duplicates(values):
+    counts = collections.Counter(values)
+    return tuple(
+        item
+        for item, num in counts.items()
+        if num > 1
+    )
+
+
+def validate_args_and_kwargs(args, kwargs, arg_names, allow_missing=False):
```
jannikluhn:
```suggestion
def validate_args_and_kwargs(args, kwargs, arg_names):
```
`allow_missing` (here as well as in `merge_kwargs_to_args`) doesn't seem to be used

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    duplicate_arg_names = _get_duplicates(arg_names)
+    if duplicate_arg_names:
+        raise TypeError("Duplicate argument names: {0}".format(sorted(duplicate_arg_names)))
```
jannikluhn: This should probably be a `ValueError`: `arg_names` is just information needed to do the validation, it's not something that's supposed to be validated itself. So duplicates there indicate an error in how the function is used.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+
+    needed_kwargs = arg_names[len(args):]
```
jannikluhn:
```suggestion
    needed_arg_names = set(arg_names[len(args):])
```
`needed_kwargs` is slightly misleading as `kwargs` are usually `dict`s. The `set` is not strictly necessary, but I'd expect it `needed_kwargs` to have the same type as `used_kwargs`.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    used_kwargs = set(arg_names[:len(args)])
```
jannikluhn:
```suggestion
    used_arg_names = set(arg_names[:len(args)])
```

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+
+    duplicate_kwargs = used_kwargs.intersection(kwargs.keys())
```
jannikluhn:
```suggestion
    duplicate_arg_names = used_arg_names.intersection(kwargs.keys())
```

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    if duplicate_kwargs:
+        raise TypeError("Duplicate kwargs: {0}".format(sorted(duplicate_kwargs)))
+
+    unknown_kwargs = set(kwargs.keys()).difference(arg_names)
```
jannikluhn:
```suggestion
    unknown_arg_names = set(kwargs.keys()).difference(arg_names)
```

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    if unknown_kwargs:
+        raise TypeError("Unknown kwargs: {0}".format(sorted(unknown_kwargs)))
+
+    missing_kwargs = set(needed_kwargs).difference(kwargs.keys())
```
jannikluhn:
```suggestion
    missing_arg_names = set(needed_kwargs).difference(kwargs.keys())
```

ssz/sedes/serializable.py
jannikluhn: I don't really like the name of the function as it's not validating the arguments itself, it's just checking that the function is called properly. But I don't have a better suggestion, so I'm fine with leaving it like this.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    if not allow_missing and missing_kwargs:
+        raise TypeError("Missing kwargs: {0}".format(sorted(missing_kwargs)))
+
+
+@to_tuple
+def merge_kwargs_to_args(args, kwargs, arg_names, allow_missing=False):
```
jannikluhn: I'm a bit surprised that there's both `merge_kwrags_to_args` and `merge_args_to_kwargs`, I suspect we can get rid of one. Working with `kwargs` feels safer, so I'd vote to keep `merge_args_to_kwargs`.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+    yield from kwargs.items()
+    for value, name in zip(args, arg_names):
+        yield name, value
+
+
+class BaseSerializable(collections.Sequence):
+    def __init__(self, *args, **kwargs):
+        if kwargs:
+            field_values = merge_kwargs_to_args(args, kwargs, self._meta.field_names)
+        else:
```
jannikluhn: I think we should be able to get rid of this, the `merge_...` functions should be able to handle empty `kwargs`, right?

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+            field_values = args
+
+        if len(field_values) != len(self._meta.field_names):
```
jannikluhn: It's not clear what validation is done where. I think it would be best to only to minimal validation in the merge function (just enough to ensure that no keyword arguments are overridden by args). Then we can have another function `validate_kwargs` (ideally with a better name) that checks that does the rest.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+            else:
+                # This is just a vanilla subclass of a `Serializable` parent class.
+                parent_serializable = serializable_bases[0]
+
+                if hasattr(parent_serializable, '_meta'):
+                    fields = parent_serializable._meta.fields
+                else:
+                    # This is a subclass of `Serializable` which has no
+                    # `fields`, likely intended for further subclassing.
+                    return super_new(cls, name, bases, attrs)
```
jannikluhn: I don't know how to fix it and it's not critical, but I'd like to point out that this line seems to be the root of this issue: https://github.com/ethereum/pyrlp/issues/101

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,419 @@
...
+        else:
+            # ensure that the `fields` property is a tuple of tuples to ensure
+            # immutability.
+            fields = tuple(tuple(field) for field in attrs.pop('fields'))
+
+        # split the fields into names and sedes
+        field_names, sedes = zip(*fields)
```
jannikluhn: This would fail if `fields` is empty. See https://github.com/ethereum/pyrlp/issues/102 for the fix.

tests/core/test_serializable.py
```diff
@@ -0,0 +1,500 @@
...
+        ('field1', uint32),
+        ('field2', uint32),
+        ('field3', uint32),
+    ]
+
+    def __init__(self, field2, field1, field3, **kwargs):
+        super().__init__(field1=field1, field2=field2, field3=field3, **kwargs)
+
+
+class SSZType4(SSZType3):
```
jannikluhn: Can we add another test class with two `uint32` fields? Then we can have a test to check that this one cannot deserialize `SSZType3` and the other way around (related to the potential container end index bug I mentioned in the comment regarding `Serializable.deserialize_segment`).

ssz/sedes/serializable.py
Bhargavasomu: Yeah, this is actually taken care by `merge_kwargs_to_args`.

ssz/sedes/serializable.py
Bhargavasomu: I think `merge_kwargs_to_args` function is needed because it takes in both `args` and `kwargs` and returns the `args` form of them combined. Like some of the fields may be given as `args`, whereas the rest maybe given as `kwargs`, but this function takes all of them and returns the `args` form of them.

ssz/sedes/serializable.py
Bhargavasomu: I will remove the `validate_args_and_kwargs` function from the `merge_kwargs_to_args` function, as it would be readable to a better extent.

ssz/sedes/serializable.py
Bhargavasomu: Fixed

ssz/sedes/serializable.py
Bhargavasomu: Fixed

tests/core/test_serializable.py
Bhargavasomu: It's working fine after I added support for classes which are declared with `empty fields` (reference https://github.com/ethereum/pyrlp/issues/101). 

Attaching the screenshot for reference.
screenshot from 2018-12-24 11-34-23

ssz/sedes/serializable.py
Bhargavasomu: It's working fine after I added support for classes which are declared with `empty fields` (reference https://github.com/ethereum/pyrlp/issues/102). 

Attaching the screenshot for reference.
screenshot from 2018-12-24 11-34-23

Bhargavasomu: Sorry, just declared the class but didn't instantiate it. Pushed another commit for that.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,425 @@
...
+            for field_name, field_sedes in cls._meta.fields
+        )
+
+        if len(serialized_string) >= 2 ** (CONTAINER_PREFIX_LENGTH * 8):
+            raise SerializationError(
+                f'Container too long to fit into {CONTAINER_PREFIX_LENGTH} bytes'
+                f'after serialization',
+                obj
+            )
+        serialized_len = len(serialized_string).to_bytes(CONTAINER_PREFIX_LENGTH, 'big')
```
hwwhww: 1. Could refactor to:

```python
len_serialized_string = len(serialized_string)

if len_serialized_string >= 2 ** (CONTAINER_PREFIX_LENGTH * 8):
    raise SerializationError(...)

serialized_len = len_serialized_string.to_bytes(CONTAINER_PREFIX_LENGTH, 'big')
```

2. @jannikluhn Or should we count the current `len(serialized_string)` after each `field_sedes.serialize(getattr(obj, field_name))`?

jannikluhn: > Or should we count the current len(serialized_string) after each field_sedes.serialize(getattr(obj, field_name))?

I think I prefer the way it is done now as it's easier to understand what's going on.

Bhargavasomu:
@jannikluhn could you review it now, have made all the changes requested. Thankyou.

hwwhww:
@jannikluhn @pipermerriam any other comment before I merge it?

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,425 @@
...
+                f"Can only serialize objects which are instances of {cls.__name__} class",
+                obj
+            )
+
+        serialized_string = b"".join(
+            field_sedes.serialize(getattr(obj, field_name))
+            for field_name, field_sedes in cls._meta.fields
+        )
+
+        if len(serialized_string) >= 2 ** (CONTAINER_PREFIX_LENGTH * 8):
```
pipermerriam: We should probably set `2 ** (CONTAINER_PREFIX_LENGTH * 8)` to a module level constant so we don't incur the overhead of computing it everytime.

ssz/sedes/serializable.py
```diff
@@ -0,0 +1,425 @@
...
+        all_kwargs = merge_args_to_kwargs(args, combined_kwargs, self._meta.field_names)
+        return type(self)(**all_kwargs)
+
+    def __copy__(self):
+        return self.copy()
+
+    def __deepcopy__(self, *args):
+        return self.copy()
+
+    _in_mutable_context = False
```
pipermerriam: I'm not seeing any code paths where this can be `True` ... am I missing something?

Bhargavasomu: You are right, I guess we aren't using this in a `mutable` context anywhere. @hwwhww could you please confirm this whether the `container datastructure` is needed to be `mutable` anywhere?

Bhargavasomu: @hwwhww I have pushed another commit assuming that it was to be `Immutable`. If this is not the case, I will revert back. Please let me know. Thankyou.

Bhargavasomu:
@pipermerriam @hwwhww ready for a final round of review hopefully.

hwwhww:
@jannikluhn I'm going to merge this now. Let us know if there's any post-merge comment.

PR #19, opened 2018-12-24
[PoC, WIP] Add BaseSSZSedes

### LINKED ISSUE #9: Introduce abstract base class for sedes objects

ChihChengLiang:
## What was wrong?

This is an attempt to fix #9

## How was it fixed?

- Naively let every sedes classes inherit abstract class BaseSSZSedes.

#### Cute Animal Picture

### DIFF
--- a/ssz/sedes/base.py
+++ b/ssz/sedes/base.py
@@ -1,0 +1,15 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
+
+
+class BaseSSZSedes(ABC):
+
+    @abstractmethod
+    def serialize(self):
+        raise NotImplementedError("SSZ Sedes classes must implement this property")
+
+    @abstractmethod
+    def deserialize(self):
+        raise NotImplementedError("SSZ Sedes classes must implement this property")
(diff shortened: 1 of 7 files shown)


Bhargavasomu:
@ChihChengLiang in the `BaseSSZSedes` class, there has to be an `abstract method` declaration for `deserialize_segment` function also.

ChihChengLiang:
Oh ... thanks! Will add them.

ChihChengLiang:
Closed in favor of #22

Issue #8, opened 2018-12-07
Add type hints


jannikluhn:
We should add type hints and check them in CI.

Bhargavasomu:
I will take this up once #17 is merged.

jannikluhn:
@Bhargavasomu Note that most of this should be done already in #22.

hwwhww:
@Bhargavasomu would you like to take it + enable mypy in tox?

Bhargavasomu:
@hwwhww I will take this up.

jannikluhn:
There's not a lot of documentation and it's unstable, but if we want to have type hints for `Serializable.__init__` (which I think we should), we probably have to write a [mypy plugin](https://mypy.readthedocs.io/en/latest/extending_mypy.html#extending-mypy-using-plugins).

pacrob:
There are a lot of types throughout the lib, but it doesn't appear that `mypy` checks were ever run as part of CI. With the incoming template updates, `mypy` is being added as part of the `pre-commit` checks, but with the `ssz` folder excluded. As files are typed, they can be added to be checked. See `pre-commit` [docs](https://pre-commit.com/#pre-commit-configyaml---hooks)

PR #22, opened 2018-12-27
Base sedes class(es)

### LINKED ISSUE #9: Introduce abstract base class for sedes objects

jannikluhn:
This PR mainly adds an abstract base class for sedes objects. In the process, the code has been restructured to avoid repetitions, introducing two more base classes.

It is fully type annotated, so most of the heavy lifting towards #8 should be completed. Therefore, I also got rid of unnecessary runtime type checks (that's what mypy is supposed to handle).

I also removed the meta information in the exceptions which is a pattern that had been inherited from pyrlp but is not used in py-evm.

This does not yet handle containers and it should probably only be merged after #17.

Closes #9.

#### Cute Animal Picture


ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+
+T = TypeVar("T")
+S = TypeVar("S")
+
+
+class BaseSedes(abc.ABC, Generic[T, S]):
+
+    @abc.abstractmethod
+    def serialize(self, value: T) -> bytes:
+        raise NotImplementedError(
```
hwwhww: I remember a wise man told me that we don't need `raise NotImplementedError()` if we use `abstractmethod`. Just `pass`.

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
+import abc
```
hwwhww: As you use `@abc.abstractmethod` a lot, maybe just import it.

```suggestion
from abc import (
    ABC,
    abstractmethod,
)
```

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @abc.abstractmethod
+    def serialize(self, value: T) -> bytes:
+        raise NotImplementedError(
+            "The serialize method must be implemented by sedes objects"
+        )
+
+    @abc.abstractmethod
+    def deserialize_segment(self, data: bytes, start_index: int) -> Tuple[S, int]:
+        raise NotImplementedError(
```
hwwhww: ditto.

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+
+    #
+    # Serialization
+    #
+    def serialize(self, value: T) -> bytes:
+        return self.serialize_content(value)
+
+    @abc.abstractmethod
+    def serialize_content(self, value: T) -> bytes:
+        raise NotImplementedError(
```
hwwhww: ditto.

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+    #
+    # Deserialization
+    #
+    def deserialize_segment(self, data: bytes, start_index: int) -> Tuple[S, int]:
+        content, continuation_index = self.consume_bytes(data, start_index, self.length)
+        return self.deserialize_content(content), continuation_index
+
+    @abc.abstractmethod
+    def deserialize_content(self, content: bytes) -> S:
+        raise NotImplementedError(
```
hwwhww: ditto.

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+    # Serialization
+    #
+    def serialize(self, value: T) -> bytes:
+        content = self.serialize_content(value)
+        self.validate_content_length(content)
+        return self.get_length_prefix(content) + content
+
+    @abc.abstractmethod
+    def serialize_content(self, value: T) -> bytes:
+        raise NotImplementedError(
```
hwwhww: ditto

ssz/sedes/base.py
```diff
@@ -0,0 +1,149 @@
...
+)
+
+from ssz.exceptions import (
+    DeserializationError,
+    SerializationError,
+)
+
+
+T = TypeVar("T")
+S = TypeVar("S")
```
hwwhww: What are `T` and `S` standing for?

jannikluhn: `T` and `S` are used because they seem to be the standard names for type variables: https://docs.python.org/3/library/typing.html#user-defined-generic-types

My guess is that `T` stands for `Type` and `S` is just the letter that is closest to `T` .

jannikluhn:
Containers and `BaseSedes` are surprisingly difficult to bring together. The reason is that the interface defined by `BaseSedes` must be satisfied by a class (`Serializable`) and not by an object. So what I thought we had to do is to let the metaclass `SerializableBase` inherit from `BaseSedes` (or rather `LengthPrefixedSedes`). This seems to work in principle, but Python behaves weird when doing `isinstance` checks then:

```Py
class Interface(ABC):

    @abstractmethod
    def method(self):
        pass

class MetaClass(type, Interface):

    def method(cls):
        print("method")

class Class(metaclass=MetaClass):
    pass

isinstance(Class(), Class)  # works
isinstance(Class, Interface)  # works
isinstance(5, Interface)  # fails with a `TypeError` in `abc.py`
isinstance(5, Metaclass)  # fails with another `TypeError` in `abc.py`
```

Alternatively, we can use virtual subclasses to trick the `isinstance` checks:

```Py
class Interface(ABC):

    @abstractmethod
    def method(self):
        pass

class MetaClass(type):

    def method(cls):
        print("method")

class Class(metaclass=MetaClass):
    pass

Interface.register(MetaClass)
```

This works, but it means we have to reimplement all (de)serialization code defined in `LengthPrefixedSedes`. So the only solution I can see right now that seems somewhat satisfying is to copy all the used methods from `LengthPrefixedSedes` in `MetaClass.__new__`, like this:

```Py
class Interface(ABC):

    @abstractmethod
    def method(self):
        pass

    def method2(self):
        print("defined in interface")
        self.method()

class MetaClass(type):

    def __new__(cls, name, bases, attrs):
        attrs["method2"] = classmethod(Interface.method2)
        return super().__new__(cls, name, bases, attrs)

    def method(cls):
        print("method")

class Class(metaclass=MetaClass):
    pass

Interface.register(MetaClass)
```

But I'm not really happy with that either as we'd have to change the metaclass whenever we add a method to `LengthPrefixedSedes`. Also, mypy probably has no idea what's going on and we won't have any type safety (that might be inevitable anyway though). Does anyone have any other ideas, maybe @pipermerriam or @cburgdorf ?

ssz/sedes/base.py
cburgdorf: Sometimes, to give type vars more meaning we also name them like `TValue`, `TKey`, `TWhatever` as a convention borrowed from the .NET community ;)

ssz/sedes/base.py
cburgdorf: Correct, `ABC` ensures classes can not be instantiated with abstract methods so the exception will never be hit and a `pass` is just a bit shorter to type.

Bhargavasomu:
@jannikluhn before you submitted this PR, I tried to finish this, and from the work I did, this is the approach I followed. It worked fine. Please take a look and let me know if this won't work. The approach is as follows.
* The class `BaseSerializable` can inherit `BaseSedes` class.
* Further in the function `is_sedes` we can modify it as follows
```
def is_sedes(obj):
    """ 
    Check if `obj` is a sedes object.
    A sedes object must inherit the class `BaseSedes`
    """
    # The issubclass check is needed for Serializable Class sedes objects
    return isinstance(obj, BaseSedes) or issubclass(obj, BaseSedes)

```

jannikluhn:
I guess this would also work, but the point is to get rid of `is_sedes` altogether and be to able to just use `isinstance(thing, BaseSedes)` instead. There should be no special case for `Serializable`s necessary.

ssz/sedes/base.py
jannikluhn: Hmm, I guess we could be more explicit here as well and get rid of the single letter variable. Maybe `TSerializable` and `TDeserialized`.

ssz/sedes/list.py
```diff
@@ -42,80 +59,53 @@ def __init__(self, element_sedes=None, empty=False):
...
-        return deserialized_data
+        if element_start_index > len(content):
+            raise Exception("Invariant: must not consume more data than available")
 
 
 address_list = List(address)
 boolean_list = List(boolean)
 bytes_list = List(bytes_sedes)
-empty_list = List(empty=True)
+empty_list: List[None, None] = List(empty=True)
```
pipermerriam: Are `None, None` the proper types here? I would have thought they'd still be something like `Sequence` or `Iterable` types but maybe I'm not understanding something.

jannikluhn: The type hint is for the element, not for the whole list (as lists in SSZ are required to have the same type for each element). The empty list has no element, so it shouldn't matter, therefore I just picked `None`.

jannikluhn:
I found a design that I'm happy with: There's a new class that handles serialization and deserialization of containers (`Container`) that's instantiated with the container fields (similar to how `List` is instantiated with `element_sedes`). The metaclass `SerializableBase` creates such a container once for each `Serializable`. It provides implementations for all methods defined in `BaseSedes` that more or less just pass through to its container instance. Finally, `SerializableBase` is made a virtual subclass of `BaseSedes`, such that each `Serializable` is an instance of `BaseSedes`.

hwwhww:
Wow the benchmark(!)

### master branch
> Performance of hash_tree_root 12.067924976348877

### This PR
First time:
> TimeoutError: hash_tree_root is not fast enough. Tolerable: 15, Actual: 29.17628002166748

Second time:
> TimeoutError: hash_tree_root is not fast enough. Tolerable: 15, Actual: 27.692071676254272

😨

jannikluhn:
> Wow the benchmark(!)
> :fearful:

I have no clue why the performance went down like this. I got it back to the original level though by moving hashing to the sedes classes (which gets rid of a lot of `isinstance` checks). However, this is an optimization that should have been possible before as well, so there might be other inefficiencies hiding somewhere.

hwwhww:
@jannikluhn It seems the performance issue is fixed in https://github.com/ethereum/py-ssz/pull/22/commits/3979d85d90a2cca521c5d7b97a74f9db55e84ca1? Although I don't understand how did that refactoring fix it. 😅

jannikluhn:
@hwwhww My best guess is that `isinstance` with virtual subclasses is slow. The commit removes the need for all instance checks, so we don't suffer from that bottleneck anymore.

hwwhww:
@jannikluhn Interesting. 😮 

BTW, I saw you added `mypy-extensitions` in dependencies, you also need to update `tox` to enable mypy checks:
```
mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p ssz
```

jannikluhn:
> BTW, I saw you added `mypy-extensitions` in dependencies, you also need to update `tox` to enable mypy checks:

Mypy is not passing yet, there are still a few places were type hints are missing or wrong that are unrelated to this PR.

@pipermerriam @cburgdorf This should be ready now, would you like to review it again or should I just merge it?

cburgdorf:
@jannikluhn nothing stands out to me. I'm :+1: to merge this (especially considering that the library is young and changing at a faster pace)

ssz/sedes/base.py
pipermerriam: This could still get cleaned up and I think more expressive names for these would improve readability.

ssz/sedes/container.py
```diff
@@ -0,0 +1,74 @@
...
+AnyTypedDict = TypedDict("AnyTypedDict", {})
+TAnyTypedDict = TypeVar("TAnyTypedDict", bound=AnyTypedDict)
+
+
+class Container(LengthPrefixedSedes[TAnyTypedDict, Dict[str, Any]]):
+
+    length_bytes = 4
+
+    def __init__(self, fields: Sequence[Tuple[str, BaseSedes[Any, Any]]]) -> None:
+        field_names = [field_name for field_name, field_sedes in fields]
```
pipermerriam: This will result in `self.fields` being mutable which doesn't seem desired.  Convert to `tuple`?

ssz/sedes/serializable.py
```diff
@@ -416,6 +338,35 @@ def __new__(cls, name, bases, attrs):
             ),
         )
 
+    #
+    # Implement BaseSedes methods as pass-throughs to the container sedes
+    #
+    def serialize(cls: Type[TSerializable], value: TSerializable) -> bytes:
```
pipermerriam: Are these intended to be `classmethod` decorated?  If not, I'm confused as to what is going on here.

jannikluhn: Those are methods defined in the metaclass, so they become classmethods automatically.

PR #24, opened 2019-01-09
Fix #23, support container type

### LINKED ISSUE #23: Support container type for tree hashing
## What was wrong?

Tree Hash should be able to hash container types.

## How can it be fixed?

- The implementation should look like this: https://github.com/ethereum/py-ssz/compare/master...ChihChengLiang:time_test?expand=1#diff-e1f1b447c1c4d61a32e846568f901c8e
- Address this issue when #17 is merged

ChihChengLiang:
## What was wrong?

Issue #23

## How was it fixed?
 
- Support container
- Support Bytes
- Add benchmarks to check if the implementation makes sense.
- Add tests, not sure how many tests are adequate to cover container types.

#### Cute Animal Picture


ssz/tree_hash/tree_hash.py
```diff
@@ -30,19 +32,38 @@
 
 
 def hash_tree_root(input_object: Any, sedes: Any=None) -> Hash32:
+    result = _hash_tree_root(input_object, sedes)
+    if len(result) < 32:
+        return Hash32(result.ljust(32, b'\x00'))
+    return Hash32(result)
+
+
+def _hash_tree_root(input_object: Any, sedes: Any=None) -> bytes:
```
jannikluhn: To me it seems like it's not worth separating this into two functions as `ljust` is only executed in one branch.

ssz/tree_hash/tree_hash.py
```diff
@@ -30,19 +32,38 @@
...
     if sedes is None:
         sedes = infer_sedes(input_object)
 
-    if isinstance(sedes, (Boolean, Hash, UnsignedInteger)):
-        serialization = sedes.serialize(input_object)
-        if isinstance(sedes, UnsignedInteger) and sedes.num_bytes > 32:
-            return hash_eth2(serialization)
+    if isinstance(sedes, (Boolean, Hash)):
```
jannikluhn: I would feel a little bit safer if each `if` would be an `elif` (with an `else` for the exception at the end). This way, if one forgets a return statement, only an exception is raised instead of a wrong value being returned.

jannikluhn:
In the long run, I think it would make sense to implement tree hashing in each sedes object (in a `sedes.tree_hash(value)` method). Then we could get rid of the `isinstance` branching. But this can only be done once #22 is merged as there are a lot of changes.

ssz/tree_hash/tree_hash.py
ChihChengLiang: Since for the integer type, the intermediate result is not ljusted while the end result is. Separating hash_tree_root into two functions is one way to tell when to use the end result.

ssz/tree_hash/tree_hash.py
```diff
@@ -30,19 +32,34 @@
...
+        container_hashes = (
+            _hash_tree_root(input_object[field_name], field_sedes)
+            for field_name, field_sedes in input_object._meta.fields
+        )
+        return hash_eth2(b''.join(container_hashes))
+    elif isinstance(input_object, abc.Iterable):
+        input_items = tuple(_hash_tree_root(item, sedes.element_sedes) for item in input_object)
+        return merkle_hash(input_items)
+    else:
+        raise TreeHashException("Can't produce tree hash", input_object, sedes)
```
hwwhww: Could you add a test case that triggers `TreeHashException`? Perhasp `tests/core/tree_hash/test_tree_hash.py`

tests/core/tree_hash/test_containers.py
```diff
@@ -0,0 +1,100 @@
...
+        (_type_1_b, SSZType1),
+        (_type_2, SSZType2),
+        (_type_3, SSZType3),
+        (_type_6, SSZType6),
+    ),
+)
+def test_serializables(value, sedes):
+
+    assert len(hash_tree_root(value, sedes)) == 32
+    # Also make sure infer works
```
hwwhww: nitpicks:
```suggestion
    # Also make sure `infer_sedes` works
```

tests/core/tree_hash/test_containers.py
```diff
@@ -0,0 +1,100 @@
...
+        ('field1', uint32),
+        ('field2', uint32),
+        ('field3', uint32),
+    ]
+
+    def __init__(self, field2, field1, field3, **kwargs):
+        super().__init__(field1=field1, field2=field2, field3=field3, **kwargs)
+
+
+class SSZType5(Serializable):
```
hwwhww: It seems another special case along with `SSZUndeclaredFieldsType`. What's different between it and `SSZUndeclaredFieldsType`? What do you think about giving it a meaningful name for describing this case?

ChihChengLiang: Sorry, I tried giving TypeX some meaningful names, but then the code becomes lengthy and variables become indistinguishable.

ChihChengLiang: I also removed type5 and type6 as these test cases are less meaningful to tree hash.

Issue #31, opened 2019-01-22
SSZ spec missing explicit checks container decoding


ChihChengLiang:
## What was wrong?

- See https://github.com/ethereum/eth2.0-specs/issues/479 for reproducible cases
- See https://github.com/ethereum/eth2.0-specs/pull/481 for the to-be-updated spec

PR #32, opened 2019-01-22
Fix 31

jannikluhn:
## What was wrong?

Issue #31

## How was it fixed?

Raise an exception in `Container.deserialize_content` if there are extra bytes at the end.

Also, fix the test we already had for this.

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_serializable.py
+++ b/tests/core/test_serializable.py
@@ -242,12 +242,14 @@ def test_container_deserialize_bad_values(value, sedes):
 def test_empty_fields_container_deserialize():
     with pytest.raises(DeserializationError):
         SSZType5.deserialize(SSZType6.serialize(_type_6))
+    with pytest.raises(DeserializationError):
         SSZType6.deserialize(SSZType5.serialize(_type_5))
 
 
 def test_subset_or_superset_fields_container_deserialize_bad_values():
     with pytest.raises(DeserializationError):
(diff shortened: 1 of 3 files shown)


tests/core/test_serializable.py
```diff
@@ -242,12 +242,14 @@ def test_container_deserialize_bad_values(value, sedes):
 def test_empty_fields_container_deserialize():
     with pytest.raises(DeserializationError):
         SSZType5.deserialize(SSZType6.serialize(_type_6))
+    with pytest.raises(DeserializationError):
         SSZType6.deserialize(SSZType5.serialize(_type_5))
```
ChihChengLiang: Would like to see a case that explicitly has a 2 fields container failed on parsing a 3 elements bytes. https://github.com/ethereum/eth2.0-specs/issues/479#issuecomment-456046599

jannikluhn: It's here: https://github.com/ethereum/py-ssz/blob/2bba715f9f9857b360045804bd9e968e4b97d4d5/tests/core/test_serializable.py#L250-L251 (with `uint32` instead of `bytes` though), or do you mean something else?

ChihChengLiang: Oh, yes, that should work.
I tested this line before your change, and it didn't raise an exception 😓. Good catch on line 252.

PR #35, opened 2019-01-31
Fix tree hash padding

jannikluhn:
## What was wrong?

https://github.com/ethereum/eth2.0-specs/pull/538

## How was it fixed?

https://github.com/ethereum/eth2.0-specs/pull/538

#### Cute Animal Picture

### DIFF
--- a/ssz/hash.py
+++ b/ssz/hash.py
@@ -31,7 +31,6 @@ def merkle_hash(input_items: Sequence[Any]) -> Hash32:
 
     # Store length of list (to compensate for non-bijectiveness of padding)
     data_length = len(input_items).to_bytes(32, "little")
-
     if len(input_items) == 0:
         # Handle empty list case
         chunks = (b'\x00' * SSZ_CHUNK_SIZE,)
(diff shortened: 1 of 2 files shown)


ssz/hash.py
```diff
@@ -40,10 +39,14 @@ def merkle_hash(input_items: Sequence[Any]) -> Hash32:
...
         # Build a list of chunks based on the number of items in the chunk
-        chunks = tuple(
+        chunks_unpadded = (
             b''.join(input_items[i:i + items_per_chunk])
             for i in range(0, len(input_items), items_per_chunk)
         )
+        chunks = tuple(
+            chunk.ljust(SSZ_CHUNK_SIZE, b"\x00")
+            for chunk in chunks_unpadded
+        )
```
hwwhww: My instinct tells me it's faster in one for loop:
```python
@to_tuple
def get_chunkz_new(input_items, items_per_chunk):
    for i in range(0, len(input_items), items_per_chunk):
        yield b''.join(input_items[i:i + items_per_chunk]).ljust(SSZ_CHUNK_SIZE, b"\x00")
```

(Although maybe not much performance improvement difference.)

Issue #36, opened 2019-02-05
Optimization


hwwhww:
## What was wrong?

- cache for the `encode`
    - ref: https://github.com/ethereum/pyrlp/blob/7904316f3da420d8dda214dfe8fa168a8e866cfb/rlp/codec.py#L48-L70
- cache for the roots
    - ref: https://github.com/ethereum/pyrlp/blob/f6e66129bda1833fdefbbc4132d13b1ab3d00fb6/rlp/sedes/serializable.py#L248-L254
 

## How can it be fixed?

Any other optimization ideas?

jannikluhn:
> Any other optimization ideas?

It would be a rather big change and I have no idea how much we would gain (if anything), but it would be interesting to look into [memoryviews](https://docs.python.org/3.7/library/stdtypes.html#memory-views). We do quite a bit of slicing when deserializing nested objects and with `bytes` this means a lot of unnecessary copying.

PR #37, opened 2019-02-06
Add `encode` cache

hwwhww:
## What was wrong?
#36 part 1: Add cache for `encode`.

## How was it fixed?
Basically copied the same logic from [pyrlp](https://github.com/ethereum/pyrlp/blob/7904316f3da420d8dda214dfe8fa168a8e866cfb/rlp/codec.py#L48-L70).

#### Cute Animal Picture

jumping-cute-playing-animals

### DIFF
--- a/ssz/codec.py
+++ b/ssz/codec.py
@@ -3,6 +3,7 @@ from eth_utils import (
 )
 
 from ssz.sedes import (
+    Serializable,
     sedes_by_name,
 )
 from ssz.sedes.base import (
@@ -13,7 +14,7 @@ from ssz.utils import (
 )
 
 
-def encode(value, sedes=None):
+def encode(value, sedes=None, cache=True):
     """
     Encode object in SSZ format.
     `sedes` needs to be explicitly mentioned for encode/decode
@@ -21,6 +22,18 @@ def encode(value, sedes=None):
     `sedes` parameter could be given as a string or as the
     actual sedes object itself.
     """
+    if isinstance(value, Serializable):
+        cached_ssz = value._cached_ssz
+        if sedes is None and cached_ssz:
+            return cached_ssz
+        else:
+            really_cache = (
+                cache and
+                sedes is None
+            )
+    else:
+        really_cache = False
+
     if sedes is not None:
         if sedes in sedes_by_name:
             # Get the actual sedes object from string representation
@@ -35,6 +48,10 @@ def encode(value, sedes=None):
         sedes_obj = infer_sedes(value)
 
     serialized_obj = sedes_obj.serialize(value)
+
+    if really_cache:
+        value._cached_ssz = serialized_obj
+
     return serialized_obj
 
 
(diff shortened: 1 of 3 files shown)


ssz/codec.py
```diff
@@ -13,14 +14,26 @@
...
     """
     Encode object in SSZ format.
     `sedes` needs to be explicitly mentioned for encode/decode
     of integers(as of now).
     `sedes` parameter could be given as a string or as the
     actual sedes object itself.
     """
+    if isinstance(value, Serializable):
+        cached_ssz = value._cached_ssz
+        if sedes is None and cached_ssz:
```
jannikluhn:
```suggestion
        if sedes is None and cached_ssz is not None:
```

just to explicitly avoid the case `cached_ssz == b""`

jannikluhn: Why do we check for `sedes is None` here?

hwwhww: Added some comment (copied from pyrlp): https://github.com/ethereum/py-ssz/pull/37/commits/d166d6c96deb080c73c2d13f7389bdd34c7bc8f1

Does that make sense to py-ssz as well? `infer_sedes` part seems similar.

hwwhww:
> Approved, even though I'm still not convinced that caching like this is very helpful. In practice, in which situations are we going to serialize the same object multiple times? I can only think of hashing, but if that's the only reason we could just cache the hash.

Yeah no good use case in my mind either in client side.


> Another thought: We could set `cached_ssz` in `ssz.decode` as well, then we could avoid serialization in some cases completely (e.g. we get something from the network and send it to other peers).

Oh yes just noticed that there are also cache for `decode`: https://github.com/ethereum/pyrlp/pull/95/ (and https://github.com/ethereum/pyrlp/blob/7904316f3da420d8dda214dfe8fa168a8e866cfb/rlp/codec.py#L245-L258). Will take a look.

tests/sedes/test_speed.py
```diff
@@ -0,0 +1,113 @@
...
+        validator_registry=tuple(validator_record for i in range(num_validators)),
+        shard_and_committee_for_slots=shard_committee_stubs,
+        latest_crosslinks=crosslink_record_stubs,
+    )
+    return state
+
+
+def do_test_serialize(state, rounds=100):
+    for _ in range(rounds):
+        x = encode(state, cache=True)
```
ChihChengLiang: This would not profile what we want. IIUC, we want elements in state to be cached, while this line would cache 100 round of the state.

I observed that the cache performed is too well even when elements in the state are random.
https://gist.github.com/ChihChengLiang/deaeee4ba41d1552f691c0bcd4aba000

hwwhww: > I observed that the cache performed too well even when elements in the state are random.

Because you only call `make_state` once?

ChihChengLiang: Ah, you're right. I'm sorry I misunderstood the cache. It's intended to cache the result of the `encode`, not elements in the state.

PR #38, opened 2019-02-06
hash_tree_root cache

hwwhww:
## What was wrong?

#36 part 2: add cache for tree hashing

## How was it fixed?

1. Add cache for `intermediate_tree_hash`.
2. Add a helper property `root` in `BaseSerializable`, we can call `BeaconBlock.root` to get the `hash_tree_root` result.

#### The performance result
Note that our test case is not the latest `BeaconState` on the spec, but it does have a long `validator_registry`.

- For the `2**18` validators case:
    - before caching: `11.3532` sec
    - ~~after caching: `0.4081` sec~~ <-- need to update to some more pratical test cases.
- For the `2**22` validators case (the worst case, 4M validators that Vitalik mentioned recently):
    - before caching: `186.1078` sec
    - ~~after caching: `6.7845` sec~~ ~~ <-- need to update to some more pratical test cases.

The performance of the worst case is still too bad... 🥀

#### Cute Animal Picture

bottlenose-dolphin-dolphin-tooth-mouth-50720

### DIFF
--- a/tests/tree_hash/test_speed.py
+++ b/tests/tree_hash/test_speed.py
@@ -1,8 +1,5 @@
 import time
 
-from ssz import (
-    hash_tree_root,
-)
 from ssz.sedes import (
     List,
     Serializable,
@@ -11,8 +8,9 @@ from ssz.sedes import (
     uint64,
     uint384,
 )
-
-TOLERABLE_PERFORMANCE = 15  # Seconds
+from ssz.tree_hash import (
+    hash_tree_root,
+)
 
 
 class ValidatorRecord(Serializable):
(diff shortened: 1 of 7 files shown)


hwwhww:
Sad news, we should generate new `ValidatorRecord` for each validator in `make_state`... then the performance becomes `6.0204` sec with cache for the of `2**18` case...

I will add other benchmarking cases.

tests/tree_hash/test_speed.py
```diff
@@ -0,0 +1,112 @@
...
+
+def do_test_tree_hash_not_cache(state, rounds=100):
+    for _ in range(rounds):
+        result = hash_tree_root(state, cache=False)
+    return result
+
+
+def do_test_tree_hash_cache(state, rounds=100):
+    for _ in range(rounds):
+        result = hash_tree_root(state, cache=True)
```
ChihChengLiang: Same issue here, states between rounds are cached

jannikluhn:
Closing because we have hashable containers now.

PR #42, opened 2019-02-13
Run YAML tests

jannikluhn:
Run SSZ tests from the eth2.0-tests repository in CI.

#### Cute Animal Picture

### DIFF
--- a/tests/yaml_tests/test_yaml.py
+++ b/tests/yaml_tests/test_yaml.py
@@ -24,13 +24,12 @@ def load_test_cases(filenames):
         with open(filename) as f:
             test = yaml.load(f)
 
-        for index, test_case in enumerate(test["test_cases"]):
-            yield test_case, make_test_id(filename, index)
+        for test_case in test["test_cases"]:
+            yield test_case, make_test_id(filename, test_case)
 
 
-def make_test_id(filename, test_case_index):
-    basename = os.path.basename(filename)
-    return f"{basename}[{test_case_index}]"
+def make_test_id(filename, test_case):
+    return f"{filename}:{test_case.lc.line}"
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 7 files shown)


jannikluhn:
> The tests are skipped, not sure if the lack of submodule update causes it.

:man_facepalming: you're right of course, fixed now.

tests/yaml_tests/test_yaml.py
```diff
@@ -0,0 +1,47 @@
...
+@to_tuple
+def load_test_cases(filenames):
+    yaml = YAML()
+
+    for filename in filenames:
+        with open(filename) as f:
+            test = yaml.load(f)
+
+        for index, test_case in enumerate(test["test_cases"]):
+            yield test_case, make_test_id(filename, index)
```
ChihChengLiang: One of the wonderful thing about `ruamel.yaml` is that it shows line numbers 🤯. So it would be interesting if path+line is included somewhere, and failing tests can be pinned down to the exact place in the thousand lines of yaml code in a snap.

As an example I hack the test id to show the path+line, and vscode allows to cmd+click to teleport me to the failing test.
```python
        for index, test_case in enumerate(test["test_cases"]):
            path = f"{index} {filename}:{test_case.lc.line}"
            yield test_case, make_test_id(filename, path)
```

screen shot 2019-02-16 at 5 36 58 pm

jannikluhn: That's amazing! Added it, can you check that the filename and line format is recognized by vscode (my editor/terminal doesn't support this unfortunately)?

ChihChengLiang: Can confirm the update works on my vscode.
Vim kind of supports it too http://vim.wikia.com/wiki/Open_file_under_cursor

PR #46, opened 2019-03-11
[WIP] Spec update

jannikluhn:
Implement changes in the latest spec update.

Main changes:

- remove `LengthPrefixedSedes` and `FixedSizedSedes` base classes in favor of `BasicSedes` and `CompositeSedes` (note that they don't overlap fully) and update sub classes accordingly.
- Add `Bytes` and `Tuple` class.
- Get rid of lots of `UInts` as they are explicitly not part of the spec anymore
- Make `BytesSedes` a list of bytes and `BytesN` a tuple of bytes (but with an optimized implementation that doesn't unnecessarily encodes every single byte separately)
- use the opportunity and rename `bytes_sedes` to `byte_list` and `BytesN` to `ByteTuple`
- some small cleanups I found on the way

What's left to do:
- fix, organize, and add some tests
- update test runner
- tree hashing

#### Cute Animal Picture


hwwhww:
Also, `SSZ_CHUNK_SIZE` was renamed to  `BYTES_PER_CHUNK` and changed to `32`.

jannikluhn:
Implemented tree hashing. Not really convinced that it's correct yet, as tests are still missing.

ssz/sedes/tuple.py
```diff
@@ -0,0 +1,108 @@
...
+    #
+    # Size
+    #
+    @property
+    def is_variable_length(self):
+        return self.number_of_elements > 0 and self.element_sedes.is_variable_length
+
+    def get_fixed_length(self):
+        if self.is_variable_length:
+            raise ValueError("Tuple does not have a fixed length")
```
pipermerriam: Based on this message I think we've got a naming issue.  The name *length* to me implies *how many elements are in the thing*, but this seems to be using it as *what size is the encoded thing*.  IIUC correctly, a `Tuple` represents a fixed size array of homogenous things, and thus, should always be a fixed length (number of elements).  That suggests to me we should be using the name `size` for this, something like `is_dynamic_sized`

ssz/sedes/tuple.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return self.element_sedes.get_fixed_length()
```
pipermerriam: This also seems odd that it would naively return it's element's value instead of something like `self.number_of_elements * self.element_sedes.get_fixed_length()` to scale the value up to the number of elements in the tuple.

ssz/utils.py
```diff
@@ -1,67 +1,120 @@
...
-        raise TypeError("Cannot infer list sedes for iterables that are not sequences")
+    return full_chunks + (last_chunk.ljust(CHUNK_SIZE, b"\x00"),)
 
-    else:
-        raise TypeError(f"Did not find sedes handling type {type(value).__name__}")
 
+def get_next_power_of_two(integer: int) -> int:
+    powers_of_two = (2**exponent for exponent in count(0))
+    greater_or_equal_powers = dropwhile(lambda power: power < integer, powers_of_two)
+    return first(greater_or_equal_powers)
```
pipermerriam: I think this is equivalent and basic profiling shows it to be about 32x faster.

```
In [23]: def alt_get_p2(v):
    return 2**v.bit_length()
```

ssz/utils.py
```diff
@@ -1,67 +1,120 @@
...
+
+
+def merkleize(chunks: Sequence[bytes]) -> bytes:
+    padded_chunks = pad_chunks(chunks)
+    number_of_layers = int(math.log2(len(padded_chunks))) + 1
+
+    layers = take(number_of_layers, iterate(hash_layer, padded_chunks))
+    root_layer = last(layers)
+    if not len(root_layer) == 1:
+        raise Exception("Invariant: root level of merkle tree has single element")
```
pipermerriam: Something I've been playing with lately for unpacking *single* value iterables with implicit validation is:

```python
root_layer, = last(layers)
```

This uses tuple unpacking and raises an exception if `last(layers)` isn't a length-1 iterable.  Saves the check but still blows up if the value is malformed.

ssz/utils.py
```diff
@@ -1,67 +1,120 @@
...
+
+def pad_chunks(chunks: Sequence[bytes]) -> Tuple[bytes]:
+    unpadded_number_of_chunks = len(chunks)
+    padded_number_of_chunks = get_next_power_of_two(unpadded_number_of_chunks)
+    padding = (EMPTY_CHUNK,) * (padded_number_of_chunks - unpadded_number_of_chunks)
+    return tuple(chunks) + padding
+
+
+def hash_layer(child_layer: Sequence[bytes]) -> Tuple[bytes]:
+    child_pairs = partition(2, child_layer)
```
pipermerriam: I *think* you know that `partition` drops *extra* elements but reading this code it's not clear whether this has been taken care of above this calling level.

Said differently, if callers of `hash_layer` haven't not properly ensured that `len(child_layer)` is even then this will blindly drop the odd trailing element.

I think that we should either 1) add that validation inline or 2) if we *know* that we can't afford the performance hit, then add a comment here explaining that callers of this function are responsible for properly normalizing inputs and that if they do not then bad things will happen.

ssz/sedes/tuple.py
jannikluhn: Agreed, using size for "number of bytes in a string" and length for "number of elements in (the first level of) a composite object" would be much cleaner. Will update the PR. Probably a good idea to use that terminology in the spec.

ssz/sedes/tuple.py
jannikluhn: yep, that's just plain wrong.

pipermerriam:
I think there's still an error with containers that contain dynamic sized elements:

```
../py-ssz/ssz/codec.py:71: in decode
    value = sedes.deserialize(ssz)
../py-ssz/ssz/sedes/serializable.py:350: in deserialize
    deserialized_field_dict = cls._meta.container_sedes.deserialize(data)
../py-ssz/ssz/sedes/base.py:59: in deserialize
    value, end_index = self.deserialize_segment(data, 0)
../py-ssz/ssz/sedes/base.py:173: in deserialize_segment
    return self.deserialize_content(content), continuation_index
../../python-environments/bimini/lib/python3.6/site-packages/eth_utils/functional.py:46: in inner
    return callback(fn(*args, **kwargs))
../py-ssz/ssz/sedes/container.py:89: in deserialize_content
    field_start_index,
../py-ssz/ssz/sedes/serializable.py:358: in deserialize_segment
    start_index,
../py-ssz/ssz/sedes/base.py:166: in deserialize_segment
    content_size = self.get_static_size()
../py-ssz/ssz/sedes/container.py:57: in get_static_size
    raise ValueError("Container contains dynamically sized elements")
E   ValueError: Container contains dynamically sized elements
```

jannikluhn:
Do you have an example? I tried the following and it seemed to work:

```Py
In [3]: class Test(ssz.Serializable): 
   ...:     fields = (("a", List(uint8)),) 
   ...:                                                

In [4]: ssz.encode(Test([1, 2, 3]))                    
Out[4]: b'\x07\x00\x00\x00\x03\x00\x00\x00\x01\x02\x03'

In [5]: ssz.decode(Out[4], Test)                       
Out[5]: <__main__.Test at 0x7fa8fa3007b8>
```

pipermerriam:
Example here:

https://github.com/ethereum/bimini/blob/master/tests/test_sss_vs_rlp.py

You probably have to manually install some of those dependencies like `numpy`, `tabulate`, `rlp`, and your local version of `ssz`

jannikluhn:
> Example here:
> 
> https://github.com/ethereum/bimini/blob/master/tests/test_sss_vs_rlp.py
> 
> You probably have to manually install some of those dependencies like `numpy`, `tabulate`, `rlp`, and your local version of `ssz`

Sorry for the delay, only got round to test it now. This takes a while but passes on my machine. The latest relevant commit should be this one: https://github.com/ethereum/py-ssz/pull/46/commits/6f6e3bb9a9385359d224056b0ecb2cb5fe7be66e

Are you sure you're on the latest version?

ChihChengLiang:
This might also relevant https://github.com/ethereum/eth2.0-specs/pull/794#pullrequestreview-215351174

jannikluhn:
@pipermerriam Found another bug while writing tests, so likely that was the one you encountered as well. Fixed now.

pipermerriam:
@jannikluhn I actually built #47 off of your branch.  I suspect we have overlapping changes but I made sufficient modifications that they're likely to be heavily obscured or just completely overwritten.  Anyways, I think this branch and #47 might need to be merged in some fashion if the SOS+SSZ spec update gets accepted.

jannikluhn:
This got out of hand a little, sorry for this. Wanted to finish this to finally move on to more important stuff. Most noteworthy changes are the added tests in `tests/sedes` and `tests/tree_hashing`.

I also got rid of lots of old tests and cleaned up `test_serializable`, even though it's unrelated to the rest of the PR.

Some minor fixes in the tree hashing functions (the helpers that contain most of the complexity haven't changed though).

Finally, I removed the YAML tests as they are outdated according to the latest spec (no support for uint24, etc.). 

For some reason, pytest and isort don't work on CircleCI anymore (even though they pass locally). Will look into this tomorrow, but it shouldn't affect the rest of the PR.

pipermerriam:
ONce you get everything green I'd advocate for breaking this up into smaller pieces.  Largely unreviewable as-is.

jannikluhn:
As suggested, I'm going to try to split this up into smaller pieces to make it reviewable.

Note to self: The tree hashing benchmark fails by a mile. I think (hope) this is because tree hashing of byte lists and vectors works with the bytes individually, but this should be fairly easy to optimize.

PR #47, opened 2019-03-15
SOS style layout

pipermerriam:
WIP with the SOS style layout as specified in https://github.com/ethereum/eth2.0-specs/pull/787

Depends on https://github.com/ethereum/py-ssz/pull/58

## What was wrong?

The spec for SSZ has been updated to a new serialization layout.

## How was it fixed?

Updated the serialization and deserialization code for all Composite types to comply with the spec.

Some API updates:

- `Container` sedes no longer operate on dictionaries or have any concept of *fields names*
- Converted to using `io.ByteIO` streams for some decoding rather than tracking the current content index and end index.

#### Cute Animal Picture

eglu_classic_green_keeping_the_foxes_at_bay


pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts= --showlocals --durations 10
```
pipermerriam: Need to revert this.

ssz/sedes/base.py
```diff
@@ -2,42 +2,50 @@
...
-    get_size_prefix,
+    encode_offset,
     merkleize,
     pack,
-    validate_content_size,
 )
 
 TSerializable = TypeVar("TSerializable")
-TDeserialized = TypeVar("TDeserialized")
+TCanonical = TypeVar("TCanonical")
```
pipermerriam: Need to revert this

pipermerriam:
@jannikluhn can you give this a preliminary review.  Not really a merge candidate until https://github.com/ethereum/eth2.0-specs/pull/787 has been merged.

ssz/utils.py
```diff
@@ -39,15 +38,24 @@ def get_duplicates(values):
     )
 
 
-def get_size_prefix(content: bytes) -> bytes:
-    return len(content).to_bytes(SIZE_PREFIX_SIZE, "little")
+def read_exact(num_bytes: int, stream: IO[bytes]) -> bytes:
+    data = stream.read(num_bytes)
+    if len(data) != num_bytes:
+        raise DeserializationError(f"Tried to read {num_bytes}.  Only got {len(data)} bytes")
```
```suggestion
        raise DeserializationError(f"Tried to read {num_bytes}. Only got {len(data)} bytes")
```

ssz/utils.py
```diff
@@ -39,15 +38,24 @@ def get_duplicates(values):
...
+    return data
 
 
-def validate_content_size(content: bytes) -> None:
-    if len(content) >= MAX_CONTENT_SIZE:
-        raise SerializationError(
-            f"Content size is too large to be encoded in a {SIZE_PREFIX_SIZE} byte prefix",
-        )
+def encode_offset(offset: int) -> bytes:
+    return offset.to_bytes(4, 'little')
```
```suggestion
    return offset.to_bytes(OFFSET_SIZE, 'little')
```

ssz/utils.py
```diff
@@ -39,15 +38,24 @@ def get_duplicates(values):
...
+
+
+def decode_offset(data: bytes) -> int:
+    return int.from_bytes(data, 'little')
+
+
+def s_decode_offset(stream: IO[bytes]) -> int:
+    data = read_exact(4, stream)
```
```suggestion
    data = read_exact(OFFSET_SIZE, stream)
```

ssz/sedes/list.py
```diff
@@ -1,119 +1,80 @@
...
+    def deserialize(self, data: bytes) -> Tuple[TElement, ...]:
+        if data:
+            raise DeserializationError("Cannot deserialize non-empty bytes using `EmptyList` sedes")
+        return tuple()
+
+    def hash_tree_root(self, value: Sequence[TElement]) -> bytes:
+        if len(value):
+            raise ValueError("Cannot compute trie hash for non-empty value using `EmptyList` sedes")
+        # TODO: This can just be a constant value, no need to compute at runtime
+        return merkleize(pack([]))
```
jannikluhn:
```suggestion
        return mix_in_length(merkleize(pack([])), 0)
```

ssz/sedes/base.py
```diff
@@ -104,77 +83,126 @@ def __init__(self, size: int):
...
+    def _iter_sedes(self, value: Sequence[Any]) -> Tuple[Tuple[Any, TSedes], ...]:
+        return tuple(zip(value, itertools.repeat(self.element_sedes)))
 
-    @abstractmethod
-    def deserialize_content(self, content: bytes) -> TDeserialized:
-        pass
+    #
+    # Tree hashing
+    #
+    def hash_tree_root(self, value: Sequence[Any]) -> bytes:
```
jannikluhn: It looks like this is only suitable for `Vector`, but not for `List` and it would make more sense to have this in `Vector` directly.

ssz/sedes/base.py
```diff
@@ -104,77 +83,126 @@ def __init__(self, size: int):
...
+        stream = io.BytesIO(data)
+        value = self.s_deserialize(stream)
+        extra_data = stream.read()
+        if extra_data:
+            raise DeserializationError(f"Got {len(extra_data)} superfluous bytes")
+        return value
 
     @abstractmethod
-    def serialize_content(self, value: TSerializable) -> bytes:
+    def s_deserialize(self, stream: IO[bytes]) -> TCanonical:
```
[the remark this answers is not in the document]
pipermerriam: The change to using streaming here is intended to just be a 1:1 replacement of the previous fragile approach where the various serialization methods returned the new position in the data where subsequent fields should read from.  I posit that these two approaches are equivalent with the stream based method being less error prone since it doesn't require manual bookkeeping to track the current position in the data.

Here is an example of the historical code:

https://github.com/ethereum/py-ssz/blob/f9afed6439046e417bbdc399eaf3ade5481b7d6c/ssz/sedes/base.py#L130-L132

Using streams, this becomes:

```python
def deserialize_segment(self, stream: IO[bytes]) -> TDeserialized:
        content = stream.read(self.size)
        return self.deserialize_content(content)
```

The new stream based decoding API is intended to be private but I realize I didn't name it as such so I'll update the name to be `_deserialize_stream` or something similar.

ssz/sedes/base.py
```diff
@@ -104,77 +83,126 @@ def __init__(self, size: int):
     #
     # Size
     #
-    @property
-    def is_static_sized(self):
-        return True
+    is_fixed_sized = True
```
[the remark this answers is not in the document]
pipermerriam: I don't think it's a concern.  I know historically I may have advocated for this safety, and I'm not opposed to adding it, but since there's nowhere in the codebase that assigns to this property I don't see it being something that's likely to go wrong.  Happy to change it back if someone prefers it to be done with `@property`

ssz/sedes/container.py
```diff
@@ -1,101 +1,142 @@
...
-        self.field_sedes_objects = tuple(field_sedes for _, field_sedes in self.fields)
-        self.field_name_to_sedes = dict(self.fields)
 
-        duplicate_field_names = get_duplicates(self.field_names)
-        if duplicate_field_names:
-            raise ValueError(
-                f"The following fields are duplicated {','.join(sorted(duplicate_field_names))}"
-            )
+class Container(CompositeSedes[Sequence[Any], Tuple[Any, ...]]):
+    def __init__(self, field_sedes: Sequence[TSedes]) -> None:
```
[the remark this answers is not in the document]
pipermerriam: Cool, I wasn't sure how this would be received but I felt it was way cleaner.

ssz/sedes/container.py
```diff
@@ -1,101 +1,142 @@
...
+        assert all(isinstance(sedes, BaseSedes) for sedes in field_sedes)
```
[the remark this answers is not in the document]
pipermerriam: This was left in on accident.  I kept hitting old code that was still passing in a dictionary and it wouldn't error out until the sedes was used so I added this to force it to error early.  Removing.

ssz/sedes/container.py
```diff
@@ -1,101 +1,142 @@
...
+        return tuple(zip(value, self.field_sedes))
+
+    def _validate_serializable(self, value: Sequence[Any]) -> bytes:
+        if len(value) != len(self.field_sedes):
+            raise SerializationError(
+                f"Not enough elements.  Expected {len(self.field_sedes)}, Got {len(value)}"
+            )
+
+    @to_tuple
+    def deserialize_section_0(self, stream: IO[bytes]) -> Iterable[Any]:
```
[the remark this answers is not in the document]
pipermerriam: I'll update to have it return two separate tuples.

ssz/sedes/base.py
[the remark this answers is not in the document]
jannikluhn: Got it, now I agree that using streams is better than passing around indices.

ssz/sedes/base.py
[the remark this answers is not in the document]
jannikluhn: No, to be honest I don't care too much either way.

pipermerriam:
@jannikluhn I addressed everything with the following exceptions.

1. You suggested trying to combine things into a single generic `deserialize` and have subclass hooks.  I'm :-1: on that approach.  Each of the different algorithms have sufficient differences that I don't see combining them as being an improvement because I feel it would decreased readability with very marginal reduction in code duplication.
2. You didn't like the mixing of values and offsets in Container deserialization.  I segregated this into a small stand-alone function and have updated it so that the separation of values from offsets is no longer happening inline.  I tinkered with an approach that didn't store them alongside each other at any point, but it added complexity at what I assessed to be no added safety since the segregation happens based on the `sedes.is_fixed_size` flag and any mechanism I thought up to keep them separate was still ultimately depending on that flag.

ssz/sedes/container.py
```diff
@@ -1,101 +1,142 @@
...
-            for field_name, field_sedes in self.fields
+    def _iter_sedes(self,
+                    value: Sequence[Any],
+                    ) -> Tuple[Tuple[Any, TSedes], ...]:
+        return tuple(zip(value, self.field_sedes))
+
+    def _validate_serializable(self, value: Sequence[Any]) -> bytes:
+        if len(value) != len(self.field_sedes):
+            raise SerializationError(
+                f"Not enough elements.  Expected {len(self.field_sedes)}, Got {len(value)}"
```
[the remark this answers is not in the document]
jannikluhn: Was this resolved on purpose?

ssz/sedes/list.py
jannikluhn: Was this resolved on purpose?

ssz/utils.py
jannikluhn: Was this resolved on purpose?

pipermerriam:
I'll update those.  I think I force pushed over top of the commits that were made by the github UI in accepting your recommended changes.

pipermerriam:
@jannikluhn I've revived the fixes that were squashed in my force pushes.

I believe this is ready for final review and merge once any issues have been addressed.

hwwhww:
@pipermerriam 


For syncing the spec versioning, the procedure we need:
1. Merge #57, #58 
2. Release `py-ssz` (sync with spec v0.5.1)
3. Merge this PR (#47) (spec dev branch)

ssz/sedes/base.py
```diff
@@ -104,77 +85,109 @@ def __init__(self, size: int):
...
-    # Deserialization
-    #
-    def deserialize_segment(self, data: bytes, start_index: int) -> Tuple[TDeserialized, int]:
-        content, continuation_index = self.consume_bytes(data, start_index, self.size)
-        return self.deserialize_content(content), continuation_index
+class BaseCompositeSedes(BaseSedes[TSerializable, TDeserialized]):
+    pass
+
+
+def _compute_section_0_size(element_sedes: Iterable[TSedes]) -> int:
```
hwwhww: IMO `section_0` and `section_1` are confusing names. `fixed_parts` and `variable_parts` from the current https://github.com/ethereum/eth2.0-specs/pull/787 PR are better.

ssz/sedes/list.py
```diff
@@ -9,112 +10,120 @@
...
+                first_offset = s_decode_offset(stream)
+            except DeserializationError:
+                # Empty list
+                return
 
-            if next_element_start_index <= element_start_index:
-                raise Exception("Invariant: must always make progress")
-            element_start_index = next_element_start_index
+            num_remaining_offset_bytes = first_offset - stream.tell()
+            if num_remaining_offset_bytes % 4 != 0:
```
hwwhww: As @jannikluhn mentioned: https://github.com/ethereum/py-ssz/pull/47/files#r272979440
Should we use `OFFSET_SIZE` instead of `4`?
```suggestion
            if num_remaining_offset_bytes % OFFSET_SIZE != 0:
```

ssz/sedes/list.py
```diff
@@ -9,112 +10,120 @@
...
+                raise DeserializationError(
+                    f"Offset bytes was not a multiple of 4.  Got {num_remaining_offset_bytes}"
```
hwwhww: ditto
```suggestion
                    f"Offset bytes was not a multiple of {OFFSET_SIZE}.  Got {num_remaining_offset_bytes}"
```

ssz/sedes/list.py
```diff
@@ -9,112 +10,120 @@
...
+                )
 
-            yield element
+            num_remaining_offsets = num_remaining_offset_bytes // 4
```
hwwhww: ditto
```suggestion
            num_remaining_offsets = num_remaining_offset_bytes // OFFSET_SIZE
```

pipermerriam:
THis has now been rebased on top of https://github.com/ethereum/py-ssz/pull/58

pipermerriam:
This PR should be ready to merge once the two dependent PR's are merged.

pipermerriam:
@jannikluhn I've rebased.  All tests are passing.  This should get one more review pass to be sure everything is fine with you and/or @hwwhww as I had to resolve a number of non-trivial merge conflicts and then resulting test failures.

ssz/sedes/base.py
```diff
@@ -102,76 +84,109 @@ def __init__(self, size: int):
...
-    def deserialize_content(self, content: bytes) -> TDeserialized:
+    def _deserialize_stream(self, stream: IO[bytes]) -> TDeserialized:
         pass
+
+
+TSerializableElement = TypeVar("TSerializableElement")
+TDeserializedElement = TypeVar("TDeserializedElement")
+
+
+class HomogenousSequence(CompositeSedes[Sequence[Any], Tuple[Any]]):
```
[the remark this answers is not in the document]
jannikluhn: Also, this class doesn't seem to be used (`Vector`, `List`, and `EmptyList` are just `CompositeSedes`).

ssz/sedes/list.py
```diff
@@ -25,95 +33,109 @@
...
     mix_in_length,
     pack,
+    read_exact,
+    s_decode_offset,
 )
 
 TSerializable = TypeVar("TSerializable")
 TDeserialized = TypeVar("TDeserialized")
 
+EMPTY_LIST_HASH_TRIES_ROOT = mix_in_length(merkleize(pack([])), 0)
```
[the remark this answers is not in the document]
pipermerriam: Keeping it here because I don't want to import `merkleize` or `pack` from within `constants.py`

ssz/sedes/list.py
```diff
@@ -25,95 +33,109 @@
...
+        if self.element_sedes.is_fixed_sized:
+            element_size = self.element_sedes.get_fixed_size()
+            data = stream.read()
+            if len(data) % element_size != 0:
+                raise DeserializationError("TODO: INVALID LENGTH")
+            for segment in partition(element_size, data):
+                yield self.element_sedes.deserialize(segment)
+        else:
+            try:
+                first_offset = s_decode_offset(stream)
```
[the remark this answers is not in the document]
pipermerriam: Fixed and added a test for this case.

pipermerriam:
@hwwhww I just realized I merged when this had "DO NOT MERGE" label on it... feel free to revert or force push master back to previous state.

hwwhww:
Whoops sorry, I need to revert it and add #62 bugfix and #60 tool into the next `py-ssz` release (mapping to spec v0.5.1).

PR #48, opened 2019-03-21
Remove unnecessary uints

jannikluhn:
Removes unsigned integers of uncommon sizes (not of the form `8 * 2**i` or larger than 32 bytes).

(originally part of #46)

#### Cute Animal Picture


jannikluhn:
With the help of [isort-seed-config](https://github.com/asottile/seed-isort-config) I finally got isort to stop complaining. It's not perfect ([this](https://github.com/ethereum/py-ssz/pull/48/files#diff-cea0ce76c7746ecb4b8221b23c2a4bc8L7) should be a new line), but at least I get consistent results now both locally and on CircleCI. Ready for review :)

PR #49, opened 2019-03-21
Implement serialization and deserialization of basic types

jannikluhn:
Adds a base class for basic types and subclasses it to implement `Boolean` as well as `UInt`. This breaks containers, lists, etc. as they expect a different base class. Therefore, all tests related to them have been removed (they're outdated anyway and will be replaced in a future PR).

Tree hashing is stubbed and will be implemented in a later PR.

Builds on top of #48, used to be part of #47.

#### Cute Animal Picture


ssz/sedes/base.py
```diff
@@ -61,24 +79,37 @@ def consume_bytes(data: bytes, start_index: int, num_bytes: int) -> Tuple[bytes,
...
+        if size <= 0:
             raise ValueError("Length must be greater than 0")
 
-        self.length = length
+        self.size = size
+
+    #
+    # Size
+    #
+    is_static_sized = True
```
hwwhww: Not sure if it's a good pattern to overwrite `@property def BaseSedes.is_static_sized()` like this.
Alternatively, maybe add `_is_static_sized` and let `is_static_sized` return `_is_static_sized`?

jannikluhn: Convinced, changed it to a property.

PR #50, opened 2019-03-21
Implement serialization and deserialization of composite types

jannikluhn:
Similar to the previous PR, add the base class `CompositeSedes` and derive `Vector`, `List`, and `Container` from it. In addition, add the necessary pass through methods to `Serializable` so that it follows the same interface as `Container`. Also contains tests.

The aliases for `ByteVector` and `ByteList` have not been implemented yet as they get their own optimized implementation that don't work with bytes directly.

Builds on top of #49, used to be part of #46.

#### Cute Animal Picture


ssz/sedes/vector.py
```diff
@@ -0,0 +1,89 @@
...
+
+TSerializableElement = TypeVar("TSerializable")
+TDeserializedElement = TypeVar("TDeserialized")
+
+
+class Vector(CompositeSedes[Sequence[TSerializableElement], Tuple[TDeserializedElement, ...]]):
+
+    def __init__(self,
+                 number_of_elements: int,
+                 element_sedes: BaseSedes[TSerializableElement, TDeserializedElement]) -> None:
```
hwwhww: Maybe reorder to `element_sedes` and then `number_of_elements`? The notation in the spec is `[type, N]`, e.g., `'latest_randao_mixes': ['bytes32', LATEST_RANDAO_MIXES_LENGTH]`.

tests/sedes/test_composite.py
```diff
@@ -0,0 +1,105 @@
...
+def test_tuple_of_dynamic_sized_entries(value, serialized):
+    sedes = Vector(len(value), List(uint8))
+    assert encode_hex(ssz.encode(value, sedes)) == serialized
+    assert ssz.decode(decode_hex(serialized), sedes) == value
+
+
+@pytest.mark.parametrize(
+    ("value", "serialized"),
+    (
+        # ((), "0x"),
```
hwwhww: It seems related to https://github.com/ethereum/eth2.0-specs/issues/854. Maybe add a TODO comment in this PR and we can come back later.

ssz/sedes/vector.py
jannikluhn: Good point. I actually ran into this a couple of times when writing the test generator.

As this changes the interface anyway, I'm going to rename `number_of_elements` to `length` as well (to be consistent terminology-wise).

tests/sedes/test_composite.py
jannikluhn: whoops, that actually works, not sure why it's commented out. Enabled it again.

PR #51, opened 2019-03-21
Aliases

jannikluhn:
Implement the aliases `Byte`, `ByteList`, and `ByteVector`. They've got their own implementation so that we can optimize them in order to not work with individual bytes.

They are used to reimplement `bytesXX` and `bytes_sedes`, which allows us to get rid of the stubs for the previously used base classes and bring back `test_speed.py`.

Builds on top of #50, used to be part of #46 

#### Cute Animal Picture


ssz/__init__.py
```diff
@@ -1,113 +1,42 @@
...
+    UInt,
+    Vector,
     boolean,
-
-    Bytes,
-    BytesN,
-    bytes_sedes,
+    byte_list,
+    byte_vector,
     bytes32,
```
hwwhww: Can I get a `bytes4` since it's used in the spec?

jannikluhn: Added `bytes64`, reread the comment, removed `bytes64`, added `bytes4`.

jannikluhn:
Unfortunately, tree hashing seems to be horribly slow. Instead of the targeted 15s the state benchmark takes about 80s on my machine. I don't really know what the issue is. Profiling suggests that most of the time is spent inside of keccak, which might mean that we don't have a lot of room for improvement.

I'm wondering though how relevant the benchmark actually is in practice. It hashes a state object with a huge number of validators. In practice however, this is not really necessary as the state only slowly grows over time and at each step we only have to calculate the updated portions of the tree (see hashable list).

Thoughts, @hwwhww @ChihChengLiang ?

ChihChengLiang:
Even running on the master branch takes 16 seconds. The fix in the validator record https://github.com/ethereum/py-ssz/pull/48/files#diff-41402ef0fe27b9b4b260ccbe4bb75774R55 reflects the real the cost of the benchmark scenario. Would suggest creating an issue for the benchmark update plan, and then address it after the rewrite.

ssz/constants.py
```diff
@@ -1 +1,5 @@
-SSZ_CHUNK_SIZE = 128
+CHUNK_SIZE = 32
```
hwwhww: Should it be renamed to `BYTES_PER_CHUNK` as the spec?

ssz/constants.py
```diff
@@ -1 +1,5 @@
...
+EMPTY_CHUNK = b"\x00" * CHUNK_SIZE
+
+SIZE_PREFIX_SIZE = 4
```
hwwhww: `BYTES_PER_LENGTH_PREFIX`?

ssz/hash.py
```diff
@@ -25,40 +25,4 @@ def hash_eth2(data: bytes) -> Hash32:
...
-        if len(chunks) % 2 == 1:
-            chunks += (b'\x00' * SSZ_CHUNK_SIZE, )
-        chunks = tuple(
-            hash_eth2(chunks[i] + chunks[i + 1])
-            for i in range(0, len(chunks), 2)
-        )
-
-    # Return hash of root and length data
-    return hash_eth2(chunks[0] + data_length)
+    return EMPTY_CHUNK
```
hwwhww: This function is not being called now, can we remove it?

ssz/sedes/byte_vector.py
```diff
@@ -0,0 +1,49 @@
+from typing import (
+    Union,
+)
+
+from ssz.exceptions import (
+    SerializationError,
+)
+from ssz.sedes.vector import (
+    CompositeSedes,
```
hwwhww: There are `ssz.sedes.base.CompositeSedes` and `ssz.sedes.vector.Vector`, I think you wanted `ssz.sedes.base.CompositeSedes`?

ssz/constants.py
jannikluhn: I tried to name everything that is a number of bytes "size" and everything that is a number of elements "length" as was suggested by Piper in the mega-PR. The spec does (or at least used to) not distinguish between the two meanings cleanly which I found confusing.

`BYTES_PER_CHUNK` would be fine on its own, but to be consistent with the rest of the code, I'd rather keep `CHUNK_SIZE`.

ssz/constants.py
jannikluhn: See above comment. Admittedly `SIZE_PREFIX_SIZE` sounds funny, but at least it should be clear what it is supposed to mean.

ssz/constants.py
jannikluhn: Added the spec constant names as comments

jannikluhn:
Rebased onto current master. Do you want to have another look @hwwhww or is your earlier approval still valid?

PR #53, opened 2019-03-21
Tree hashing

jannikluhn:
Implements tree hashing for basic, composite and (in a to be optimized version) for alias types.

Builds on top of #52, used to be part of #46 

#### Cute Animal Picture


hwwhww:
https://github.com/ethereum/py-ssz/pull/51#issuecomment-475896906
> Unfortunately, tree hashing seems to be horribly slow. Instead of the targeted 15s the state benchmark takes about 80s on my machine. I don't really know what the issue is. Profiling suggests that most of the time is spent inside of keccak, which might mean that we don't have a lot of room for improvement.

> I'm wondering though how relevant the benchmark actually is in practice. It hashes a state object with a huge number of validators. In practice however, this is not really necessary as the state only slowly grows over time and at each step we only have to calculate the updated portions of the tree (see hashable list).


FYI a quick benchmark comparing with minimal SSZ: https://github.com/hwwhww/py-ssz/commit/b78ca9dd1f0026a276ed379b0494947e45cd2a86 with `2**18` validators, and make BeaconState only contain `validator_registry`:
```
Performance of hash_tree_root 52.62116575241089
Performance of mini_hash_tree_root 35.773415327072144
```
1. IIRC, `HashableList` can help when updating validators, but it's not helpful when we do the first time tree hashing of the whole state?
2. Farther optimization (not in this PR), maybe we can implement a way to parallelly do (i) Merkle leaves or (ii) `hash_tree_root` each validator in SSZ module level?

jannikluhn:
> FYI a quick benchmark comparing with minimal SSZ

Nice, so it's actually not super bad

> IIRC, HashableList can help when updating validators, but it's not helpful when we do the first time tree hashing of the whole state?

I guess it depends on what you mean with "first time":

- new sync from genesis: The state is reconstructed and is tiny at the beginning, so that shouldn't be a problem
- new sync from "checkpoint" downloaded from peer: We should download data with proofs anyway, so we don't need to do any hashing other than verifying the proof (and bandwidth is probably the bottleneck here).
- new sync from checkpoint downloaded from trusted source (Trinity Github repo): The data could just contain the full hash tree
- restart: We should store the hash tree in a DB, so no need to rehash

> Farther optimization (not in this PR), maybe we can implement a way to parallelly do (i) Merkle leaves or (ii) hash_tree_root each validator in SSZ module level?

Interesting idea. But I think we should first try to implement `HashableList/Container` and see if we actually need it, as this would introduce some complexity.

ssz/sedes/list.py
```diff
@@ -104,5 +110,24 @@ def deserialize_content(self, content: bytes) -> Generator[TDeserialized, None,
         if element_start_index > len(content):
             raise Exception("Invariant: must not consume more data than available")
 
+    #
+    # Tree hashing
+    #
+    def hash_tree_root(self, value: Iterable[TSerializable]) -> bytes:
+        if isinstance(self.element_sedes, BasicSedes):
```
ChihChengLiang: Using `hasattr` can buy 1 sec in the benchmark, but I'm super uncertain about this.
```suggestion
        if hasattr(self.element_sedes, "serialize"):
```

ChihChengLiang: Nevermind, it's random noise

jannikluhn: I don't think this is worth it at the moment. And if it turns out that the `isinstance` checks are a major bottleneck maybe we can try to cache it somewhere before reverting to the old `hasattr` checks.

PR #56, opened 2019-04-08
Fix SerializableBase.get_static_size()

hwwhww:
## What was wrong?

Missing `return` value in `SerializableBase.get_static_size()`

We need this patch for `BeaconState`.

## How was it fixed?

Add it.

#### Cute Animal Picture

🐼


jannikluhn:
> We need this patch for BeaconState.

So shall we make a new release?

hwwhww:
@jannikluhn yes, please! :)

jannikluhn:
Done

lookfirst:
@jannikluhn There is no unit test that would have caught this?!

PR #57, opened 2019-04-08
Serializable metaclass cleanups

jannikluhn:
This started as some preparations for implementing signed roots, but ended up with some bigger cleanups of the (still a bit messy) serializable metaclass.

- make the code easier to follow in general
- renamed `SerializableBase` to `MetaSerializable` to make it clear that it's a meta class and to distinguish it from `BaseSerializable`
- change the inheritance rules: So far you had to either have only one parent with fields and inherit those, or have multiple parents and define your own fields which have to contain all fields defined by all ancestors. Now, you can replace fields with whatever you like, but have to do so whenever you have multiple parents which define fields.
- simplify and extend tests for inheritance structure
- handle empty field lists (by failing with a TypeError)

#### Cute Animal Picture


ssz/sedes/serializable.py
```diff
@@ -184,16 +194,45 @@ def _mk_field_attrs(field_names, extra_namespace):
...
 
-def _mk_field_property(field, attr):
-    def field_fn_getter(self):
-        return getattr(self, attr)
+@to_dict
+def _mk_field_props(field_names, field_attrs):
+    # we can't just define the getter in the loop as attr in the getter scope would not refer to
+    # the current value of the loop variable, but always the last one (see
+    # https://stackoverflow.com/q/233673)
+    def getter_factory(attr):
```
hwwhww: coool 😎

ssz/sedes/serializable.py
```diff
@@ -214,127 +253,103 @@ def _get_class_namespace(cls):
         yield from cls.__slots__
 
 
-class SerializableBase(abc.ABCMeta):
+class MetaSerializable(abc.ABCMeta):
 
-    def __new__(cls, name, bases, attrs):
-        super_new = super(SerializableBase, cls).__new__
+    def __new__(mcls, name, bases, namespace):
+        declares_fields = "fields" in namespace
```
hwwhww: Maybe we can set string `"fields"` as a constant? (Just worry about typo.)

ssz/sedes/serializable.py
```diff
@@ -27,11 +34,13 @@
...
-    fields = None
-    field_names = None
-    field_attrs = None
-    container_sedes = None
+class Meta(NamedTuple):
+
+    has_fields: bool
+    fields: Optional[Tuple[Tuple[str, BaseSedes]]]
+    container_sedes: Optional[Container]
+    field_names: Optional[Tuple[str]]
```
pipermerriam: `Tuple[str]` denotes a length-1 tuple with a single string.  Also, why is this `Optional`?  Shouldn't there always be field names (or an empty tuple)

pipermerriam: `Tuple[str]` denotes a length-1 tuple with a single string.   Should probably be `Tuple[str, ...]` for all of these.

ssz/sedes/serializable.py
pipermerriam: The standard library `operator.attrgetter(attr)` does exactly this.

ssz/sedes/serializable.py
```diff
@@ -214,127 +253,103 @@ def _get_class_namespace(cls):
...
-        super_new = super(SerializableBase, cls).__new__
+    def __new__(mcls, name, bases, namespace):
+        declares_fields = "fields" in namespace
 
-        serializable_bases = tuple(b for b in bases if isinstance(b, SerializableBase))
-        has_multiple_serializable_parents = len(serializable_bases) > 1
-        is_serializable_subclass = any(serializable_bases)
-        declares_fields = 'fields' in attrs
+        if declares_fields:
+            has_fields = True
```
pipermerriam: What if `len(namespace['fields'] == 0`.  Should that be taken into account here?

ssz/sedes/serializable.py
```diff
@@ -214,127 +253,103 @@ def _get_class_namespace(cls):
...
+            fields = namespace.pop("fields")
+            try:
+                sedes = Container(fields)
+            except ValueError as exception:
```
pipermerriam: Since `ValueError` is a common standard library exception maybe the source of this should be converted to an `eth_utils.ValidationError` to ensure we're catching the right thing?  Also probably worth a comment explaining why this is being caught and re-raised.

tests/misc/test_serializable_inheritance.py
```diff
@@ -0,0 +1,113 @@
...
+def create_inheritance_structure(structure):
+    class_definitions = tuple(definition.strip() for definition in structure.split())
+
+    class_dict = {}
+    field_names_dict = {}
+
+    for index, class_definition in enumerate(class_definitions):
+        # get the name of the class
+        class_name_start = 0
+        class_name_end = class_definition.index("(")
```
pipermerriam: Not necessary but it would be pretty trivial to use pyparsing to write this up as a simple grammar which would make the parsing a bit easier to understand.  Probably not needed in this PR but something I think I could do pretty quickly if we think readability of this code is important.

tests/misc/test_serializable_inheritance.py
```diff
@@ -0,0 +1,113 @@
...
+))
+def test_valid_inheritance(structure):
+    class_dict, field_names_dict = create_inheritance_structure(structure)
+
+    for class_name in class_dict.keys():
+        class_ = class_dict[class_name]
+        field_names = field_names_dict[class_name]
+
+        kwargs = {field_name: 0 for field_name in field_names}
+        class_(**kwargs)
```
pipermerriam: Might be good to add some assertions about things like each class having the declared field names.

ssz/sedes/serializable.py
jannikluhn: In this case the container initialization two lines below would fail and a `TypeError` would be raised. I've added a comment to make it more clear.

tests/misc/test_serializable_inheritance.py
jannikluhn: Good idea! Didn't know about pyparsing, it seems like exactly the right tool for the job.

Issue #44, opened 2019-02-19
Signed roots


ChihChengLiang:
## What was wrong?

https://github.com/ethereum/eth2.0-specs/pull/625

hwwhww:
closing via #58

PR #58, opened 2019-04-08
Root and SignedRoot

jannikluhn:
## What was wrong?

#44 

## How was it fixed?

- add the property `root` to `BaseSerializable`
- add a metaclass for signed serializables that validates that the last field is called signature and there's at least one more (otherwise we can't calculate the signed root, assuming empty containers don't exist)
- add a new base class `SignedSerializable` that provides the property `signed_root`

(builds on top of #57)

#### Cute Animal Picture


ssz/sedes/serializable.py
```diff
@@ -27,11 +35,13 @@
 TSerializable = TypeVar("TSerializable", bound="BaseSerializable")
 
 
-class MetaBase:
-    fields = None
-    field_names = None
-    field_attrs = None
-    container_sedes = None
+class Meta(NamedTuple):
+
```
ralexstokes: minor but i feel like we usually have no whitespace b/t class decl and class attrs

ssz/sedes/signed_serializable.py
```diff
@@ -0,0 +1,70 @@
...
+    Container,
+)
+from ssz.sedes.serializable import (
+    BaseSerializable,
+    MetaSerializable,
+)
+
+
+class SignedMeta(NamedTuple):
+    has_fields: bool
```
ralexstokes: example about formatting here

ssz/sedes/signed_serializable.py
```diff
@@ -0,0 +1,70 @@
...
+    fields: Optional[Tuple[Tuple[str, BaseSedes]]]
+    container_sedes: Optional[Container]
+    signed_container_sedes: Optional[Container]
+    field_names: Optional[Tuple[str]]
+    field_attrs: Optional[Tuple[str]]
+
+
+class MetaSignedSerializable(MetaSerializable):
+
```
ralexstokes: whitespace? i'll let you decide if you want to patch these up, not blocking merging imo

ssz/sedes/serializable.py
jannikluhn: :+1: I don't like how it looks, but I do like conventions, so I removed the new lines for all classes I could find.

PR #59, opened 2019-04-23
Use SHA256 instead of Keccak256

ralexstokes:
## What was wrong?

We have switched to use SHA256.

## How was it fixed?

Drop `eth-hash` and use the stdlib `sha256` hash function.

#### Cute Animal Picture


hwwhww:
looks good!
I labeled it as `DO NOT MERGE` for now - it will be merged with the spec v0.6.0 release.

hwwhww:
removed the `DO NOT MERGE` label!

PR #63, opened 2019-04-26
SOS style layout

hwwhww:
Repost #47


pipermerriam:
As soon as this is merged/released I'll finish getting the test generators updated.

hwwhww:
removed the `DO NOT MERGE` label!

pipermerriam:
I'll get this rebased this morning.

pipermerriam:
@hwwhww this has been rebased

ssz/sedes/container.py
```diff
@@ -1,104 +1,163 @@
...
+    def __init__(self, field_sedes: Sequence[TSedes]) -> None:
+        if len(field_sedes) == 0:
             raise ValidationError("Cannot define container without any fields")
-
-        duplicate_field_names = get_duplicates(self.field_names)
-        if duplicate_field_names:
-            raise ValidationError(
-                f"The following fields are duplicated {','.join(sorted(duplicate_field_names))}"
-            )
+        self.field_sedes = tuple(field_sedes)
```
hwwhww: Ouch, just realize that, with this change, we can't [hack vector length with `Container.field_name_to_sedes` dict directly](https://github.com/ethereum/trinity/blob/master/eth2/beacon/tools/misc/ssz_vector.py), and the field names are not available from `to_formatted_dict()` result.

1. Is there any chance to keep the field names available in `Container`? It can be `field_names: Tuple[str, ...]` instead of `field_name_to_sedes: Dict[str, Sequence[TSedes]]` if you want to make it immutable if we can still do forced update the length in Trinity tests.
2. IMO having field names in `to_formatted_dict()` result is a requirement.

pipermerriam: @hwwhww Yes, this should be updated to return a dictionary with the names intact.  I'll make that change.

pipermerriam: Delete my previous response.

I **do not** think that `Container` should  know about field names, however, I think that `Serializable` can adjust for this.

We can updated the `to_formatted_dict` implementation for `Serializables` to return a `Mapping`, preserving the field names.  Does that solve your problem?

hwwhww:
@pipermerriam 
I think that's workable to make `Serializable` have API to access field names instead of via the protected attribute `self._meta`. 👍 


The trinity tests can be updated to like:
```python
for key, value in state_vector_dict.items():
    # BeaconState._meta.container_sedes.field_sedes[key].length = value
    index = BeaconState._meta.field_names.index(key)
    BeaconState._meta.container_sedes.field_sedes[index].length = value

```

pipermerriam:
Just pushed 62c1b7b  does that address your issues?

As for the *privateness* of `Serializable._meta`, I'm not sure whether it's better to expose that or just accept that accessing `Serializable._meta` is *ok* and that it lives there to prevent easy namespace colliding with field names  (Django does this to a degree)

hwwhww:
@pipermerriam 
e8dc795 addresses my issues. 👍 

> As for the privateness of `Serializable._meta`

Hmm, IMHO I think it's *okay*. But /cc @jannikluhn

jannikluhn:
> As for the privateness of `Serializable._meta`

I also think it's fine.

pipermerriam:
I could be convinced to implement an API something like:

- `instance.get_meta('field_names')`
- `instance.get_meta(SomeEnum.FIELD_NAMES)`
- `instance.get_meta().field_names`  (probably my strongest preference)

I think that `get_meta` is less likely to collide.  We could improve that by making it more verbose like `get_serializable_meta`

PR #64, opened 2019-04-26
Fix `BaseSerializable.__eq__` bug

### LINKED ISSUE #62: `BaseSerializable.__eq__` bug
## What was wrong?

The current `BaseSerializable.__eq__` overwritten:

https://github.com/ethereum/py-ssz/blob/91f0928cd8fb8b8f3f67fc6799a31f14daf98b30/ssz/sedes/serializable.py#L137-L138

I found it's incorrect when I do the `==` assertion with `SignedSerializable` object, where `SignedSerializable`  inherits `BaseSerializable`, not `Serializable`:

https://github.com/ethereum/py-ssz/blob/91f0928cd8fb8b8f3f67fc6799a31f14daf98b30/ssz/sedes/signed_serializable.py#L65-L68

## How can it be fixed?
1. Fix `BaseSerializable.__eq__`: change `isinstance(other, Serializable)` to `isinstance(other, BaseSerializable)`
2. Question: should `SignedSerializable`  inherits `Serializable` instead of `BaseSerializable`?
3. Question: use `root` (`hash_tree_root`) instead of bulti-in `hash`? 
https://github.com/ethereum/py-ssz/blob/91f0928cd8fb8b8f3f67fc6799a31f14daf98b30/ssz/sedes/serializable.py#L140-L155

hwwhww:
## What was wrong?

Fix #62 `BaseSerializable.__eq__` bug

## How was it fixed?

1. Move `pytest-watch` to test dep (so I can test with trinity)
2. Check `isinstance(other, BaseSerializable)` instead of `isinstance(other, Serializable)`
3. Add a test for equality

#### Cute Animal Picture

easter-2197043_640

### DIFF
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -132,7 +132,7 @@ class BaseSerializable(collections.Sequence):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return isinstance(other, BaseSerializable) and hash(self) == hash(other)
+        return self.__class__ == other.__class__ and hash(self) == hash(other)
 
     def __getstate__(self):
         state = self.__dict__.copy()
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -132,7 +132,7 @@ class BaseSerializable(collections.Sequence):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return isinstance(other, Serializable) and hash(self) == hash(other)
+        return isinstance(other, BaseSerializable) and hash(self) == hash(other)
 
     def __getstate__(self):
         state = self.__dict__.copy()
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -132,7 +132,7 @@ class BaseSerializable(collections.Sequence):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return self.__class__ == other.__class__ and hash(self) == hash(other)
+        return self.__class__ is other.__class__ and hash(self) == hash(other)
 
     def __getstate__(self):
         state = self.__dict__.copy()
--- a/tests/misc/test_signed_serializable.py
+++ b/tests/misc/test_signed_serializable.py
@@ -58,3 +58,19 @@ def test_signing_root():
     signed = Signed(123, b"\xaa", b"\x00")
     unsigned = Unsigned(123, b"\xaa")
     assert signed.signing_root == unsigned.root
+
+
+def test_eq():
+    class Signed(ssz.SignedSerializable):
+        fields = (
+            ("field1", uint8),
+            ("field2", byte_list),
+            ("signature", byte_list),
+        )
+
+    signed_a = Signed(123, b"\xaa", b"\x00")
+    signed_b = signed_a.copy()
+    assert signed_a == signed_b
+
+    signed_b = signed_b.copy(field1=456)
+    assert signed_a != signed_b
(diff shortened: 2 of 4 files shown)


tests/misc/test_signed_serializable.py
```diff
@@ -58,3 +58,19 @@ class Unsigned(ssz.Serializable):
...
+            ("field2", byte_list),
+            ("signature", byte_list),
+        )
+
+    signed_a = Signed(123, b"\xaa", b"\x00")
+    signed_b = signed_a.copy()
+    assert signed_a == signed_b
+
+    signed_b = signed_b.copy(field1=456)
+    assert signed_a != signed_b
```
jannikluhn: Maybe add a test that checks that a signed serializable is equal to an unsigned serializable if all fields match.

pipermerriam: That API is surprising to me.  Seems like equality checks between those two returning `True` might violate principle of least surprise.

jannikluhn: To clarify: A signed serializable is equal to an unsigned serializable only if they have the same fields *including* the signature. This is what I would expect: The two Python objects represent the same SSZ object (same fields, same data, same root). The only difference is that the `SignedSerializable` has some additional convenience features.

hwwhww: @jannikluhn @pipermerriam Thanks for the feedback!

Added more tests: https://github.com/ethereum/py-ssz/pull/64/commits/5e32ef2bf3295a67ca2d4db7969453a3293402ac, and yep, a little surprising that signed serializable `==` to an unsigned serializable.

Should we change it to checking the same `__class__`? i.e.,:
```python
    def __eq__(self, other):
        return self.__class__ == other.__class__ and hash(self) == hash(other)
```

hwwhww: I left the previous comment before I saw Jannik's latest comment.

I suppose for the use case, we would use `==` for the fields check and `is` for distinguishing the object.

`SignedSerializable` might be a special case; another case is that do we expect a `PetersburgBlock` object `==` a `ByzantiumBlock` object?

pipermerriam: so IIUC:

```python
class Thing(Serializable):
    fields = ...

class SignedThing(SignedSerializable):
    fields = ...

thing = Thing(...)
signed_thing = SignedThing(...)  # same field values as `thing` but with signature

# we expect this to be true
assert thing == signed_thing
```

If so, then this is indeed surprising to me and doesn't seem like a good API choice.

jannikluhn: What would be the less confusing behavior be instead? I can see four options:

1) equal only if class is equal
3) equal if one class is the same or a subclass of the other
2) equal if both are `Serializable` or both are `SignedSerializable`, but not if one is signed and the other isn't
4) equal if both inherit from the same `fields` defining parent

hwwhww: I'd say (1). In the use cases, we won't have a `Thing` and a `SignedThing` with the same fields.

pipermerriam: I think 1 is the safest and easiest to relax the constraint to become 3 or 4 if the use case proves useful.

ssz/sedes/serializable.py
```diff
@@ -132,7 +132,7 @@ def __len__(self):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return isinstance(other, BaseSerializable) and hash(self) == hash(other)
+        return self.__class__ == other.__class__ and hash(self) == hash(other)
```
pipermerriam: *nitpick* since classes are singletons, `self.__class__ is other.__class__` is maybe slightly better here.

PR #66, opened 2019-04-27
Fix #65, add Formatted dict

### LINKED ISSUE #65: Import/Export to json or yaml
## What was wrong?

#60 provides basic functionality to import/export a `Serializable` from/to a json/yaml file, but it looks like there are rooms for improvement. Here are some examples:

- Provide `to_json(self)` and `to_yaml(self)` methods to `Serializable`
- The import/export should be tools and utilities, but the relevant methods are touching the core classes. Maybe need a way to separate them.
- Should be able to customize the output of Uints to digits or hex to support different use cases.

## How can it be fixed?

Requires discussion

ChihChengLiang:
## What was wrong?

Issue #65

## How was it fixed?

- Reuse the yaml test parser 
- Support `Serializable` to/from formatted dict
- Support import/export int as int/str/hex

#### Cute Animal Picture

image-compressor


hwwhww:
New error message with https://github.com/hwwhww/trinity/commit/662410bf48d5b4cc71944b1bb43e847a74225f92:
```
___________________________________________________________________________ test_state ____________________________________________________________________________

base_db = <eth.db.atomic.AtomicDB object at 0x10eff1eb8>

    def test_state(base_db):
        with open(BASE_FIXTURE_PATH + '/' + FILE_NAME, 'r') as stream:
            try:
                data = yaml.safe_load(stream)
            except yaml.YAMLError as exc:
                print(exc)

        test_cases = data['test_cases']
        for test_case in test_cases:
>           execute_state_transtion(test_case, base_db)

base_db    = <eth.db.atomic.AtomicDB object at 0x10eff1eb8>
data       = {'fork': 'v0.5.1', 'summary': 'Basic sanity checks from phase 0 spec pythonization using a small state configuration a...00000000000000000000000000', 'current_shuffling_epoch': 536870912, ...}, ...}, ...], 'test_suite': 'beacon_state', ...}
stream     = <_io.TextIOWrapper name='/Users/hwwang/ethereum/trinity/tests/eth2/../../eth2-fixtures/state/sanity-check_small-config_32-vals.yaml' mode='r' encoding='UTF-8'>
test_case  = {'blocks': [{'body': {'attestations': [], 'attester_slashings': [], 'deposits': [], 'eth1_data': {'block_hash': '0x000... '0x0000000000000000000000000000000000000000000000000000000000000000', 'current_shuffling_epoch': 536870912, ...}, ...}
test_cases = [{'blocks': [{'body': {'attestations': [], 'attester_slashings': [], 'deposits': [], 'eth1_data': {'block_hash': '0x00...00000000000000000000000000000000000000000000000000000000000000', 'current_shuffling_epoch': 536870912, ...}, ...}, ...]

tests/eth2/beacon/state-fixtures/test_minimal_state.py:63:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/eth2/beacon/state-fixtures/test_minimal_state.py:102: in execute_state_transtion
    block = SerenityBeaconBlock.from_formatted_dict(dict_block)
../py-ssz/ssz/sedes/serializable.py:191: in from_formatted_dict
    return io.parse_value(value, cls)
../py-ssz/ssz/tools/io.py:58: in parse_value
    input_dict = self.parse_value(value, sedes._meta.container_sedes)
../py-ssz/ssz/tools/io.py:95: in parse_value
    for field_name, field_value in value.items()
../py-ssz/ssz/tools/io.py:95: in <dictcomp>
    for field_name, field_value in value.items()
../py-ssz/ssz/tools/io.py:58: in parse_value
    input_dict = self.parse_value(value, sedes._meta.container_sedes)
../py-ssz/ssz/tools/io.py:95: in parse_value
    for field_name, field_value in value.items()
../py-ssz/ssz/tools/io.py:95: in <dictcomp>
    for field_name, field_value in value.items()
../py-ssz/ssz/tools/io.py:72: in parse_value
    return tuple(self.parse_value(element, sedes.element_sedes) for element in value)
../py-ssz/ssz/tools/io.py:72: in <genexpr>
    return tuple(self.parse_value(element, sedes.element_sedes) for element in value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ssz.tools.io.FormattedDictIO object at 0x10f1739e8>
value = {'header_1': {'block_body_root': '0x0000000000000000000000000000000000000000000000000000000000000000', 'previous_block...25fa38acd1646d155eeedf77c76b024c0adda0317935f298982635c6bbb047218f38a', 'slot': 4294967297, ...}, 'proposer_index': 31}
sedes = 'ProposerSlashing'

    def parse_value(self, value, sedes=None):
        if isinstance(sedes, MetaSerializable):
            serializable_cls = sedes
            input_dict = self.parse_value(value, sedes._meta.container_sedes)
            return serializable_cls(**input_dict)

        elif isinstance(sedes, Boolean):
            if not isinstance(value, bool):
                raise ValueError(f"Expected value of type bool, got {type(value)}")
            return value

        elif isinstance(sedes, UInt):
            return self._parse_int(value)

        elif isinstance(sedes, List):
            if not isinstance(value, Sequence):
                raise ValueError(f"Expected list, got {type(value)}")
            return tuple(self.parse_value(element, sedes.element_sedes) for element in value)

        elif isinstance(sedes, Vector):
            if not isinstance(value, Sequence):
                raise ValueError(f"Expected list, got {type(value)}")
            if not len(value) == sedes.length:
                raise ValueError(f"Expected {sedes.length} elements, got {len(value)}")
            return tuple(self.parse_value(element, sedes.element_sedes) for element in value)

        elif isinstance(sedes, (ByteList, ByteVector)):
            return decode_hex(value)

        elif isinstance(sedes, Container):
            if not isinstance(value, Mapping):
                raise ValueError(f"Expected mapping, got {type(value)}")
            field_names_got = set(value.keys())
            field_names_expected = set(field_name for field_name, _ in sedes.fields)
            if field_names_got != field_names_expected:
                raise ValueError(
                    f"Unexpected fields: Got {field_names_got} instead of {field_names_expected}"
                )
            return {
                field_name: self.parse_value(field_value, sedes.field_name_to_sedes[field_name])
                for field_name, field_value in value.items()
            }

        elif isinstance(sedes, BaseSedes):
            raise Exception("Unreachable: All sedes types have been checked")

        else:
>           raise TypeError("Expected BaseSedes")
E           TypeError: Expected BaseSedes

sedes      = 'ProposerSlashing'
self       = <ssz.tools.io.FormattedDictIO object at 0x10f1739e8>
value      = {'header_1': {'block_body_root': '0x0000000000000000000000000000000000000000000000000000000000000000', 'previous_block...25fa38acd1646d155eeedf77c76b024c0adda0317935f298982635c6bbb047218f38a', 'slot': 4294967297, ...}, 'proposer_index': 31}

../py-ssz/ssz/tools/io.py:102: TypeError
==================================================================== slowest 50 test durations ====================================================================
2.99s call     tests/eth2/beacon/state-fixtures/test_minimal_state.py::test_state
0.01s setup    tests/eth2/beacon/state-fixtures/test_minimal_state.py::test_state
0.00s teardown tests/eth2/beacon/state-fixtures/test_minimal_state.py::test_state
==================================================================== 1 failed in 3.23 seconds =====================================================================
```

ssz/tools/io.py
```diff
@@ -0,0 +1,152 @@
...
+                raise ValueError(f"Expected mapping, got {type(value)}")
+            field_names_got = set(value.keys())
+            field_names_expected = set(field_name for field_name, _ in sedes.fields)
+            if field_names_got != field_names_expected:
+                raise ValueError(
+                    f"Unexpected fields: Got {field_names_got} instead of {field_names_expected}"
+                )
+            d = {}
+            for field_name, field_value in value.items():
+                print(field_name)
```
hwwhww: Should remove it.
```suggestion
```

ChihChengLiang: Ahhh ... nice catch 🙈

ssz/sedes/serializable.py
```diff
@@ -179,6 +179,17 @@ def __deepcopy__(self, *args):
     def root(self):
         return ssz.hash_tree_root(self)
 
+    def to_formatted_dict(self, int_as="int"):
+        from ssz.tools.io import FormattedDictIO
```
jannikluhn: Should be at the top of the file

ssz/sedes/serializable.py
```diff
@@ -179,6 +179,17 @@ def __deepcopy__(self, *args):
...
+        io = FormattedDictIO(int_as=int_as)
+        return io.dump_value(self)
+
+    @classmethod
+    def from_formatted_dict(cls, value, int_as="int"):
+        from ssz.tools.io import FormattedDictIO
```
jannikluhn: ditto

ssz/examples/zoo.py
```diff
@@ -0,0 +1,75 @@
...
+        ('epoch', uint64),
+        ('bio_id_scan', byte_list),
+        ('signature', bytes96),
+    ]
+
+
+class Animal(Serializable):
+    fields = [
+        ('id_hash', bytes32),
+        ('public_key', bytes48),
```
jannikluhn: :joy:

tests/tools/test_io.py
```diff
@@ -0,0 +1,24 @@
...
+        "hex",
+    )
+)
+def test_import_export(tmpdir, int_as):
+    path = tmpdir / "zoo.dump"
+    with open(path, "w") as f:
+        json.dump(zoo.to_formatted_dict(int_as=int_as), f)
+    with open(path, "r") as f:
+        read_zoo = Zoo.from_formatted_dict(json.load(f), int_as=int_as)
+    assert read_zoo.animals[0].clock_in_records[0].epoch == zoo.animals[0].clock_in_records[0].epoch
```
jannikluhn: Why not

```suggestion
    assert read_zoo == zoo
```

?

tests/tools/test_io.py
```diff
@@ -0,0 +1,24 @@
...
+    "int_as", (
+        "int",
+        "str",
+        "hex",
+    )
+)
+def test_import_export(tmpdir, int_as):
+    path = tmpdir / "zoo.dump"
+    with open(path, "w") as f:
+        json.dump(zoo.to_formatted_dict(int_as=int_as), f)
```
jannikluhn: Not sure if writing it to a file is necessary here, `json.dumps` seems sufficient.

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+            return encode_hex(value)
+
+        elif isinstance(sedes, Container):
+            if not isinstance(value, Mapping):
+                raise ValueError(f"Expected mapping, got {type(value)}")
+            field_names_got = set(value.keys())
+            field_names_expected = set(field_name for field_name, _ in sedes.fields)
+            if field_names_got != field_names_expected:
+                raise ValueError(
+                    f"Unexpected fields: Got {field_names_got} instead of {field_names_expected}"
```
hwwhww: You can output `field_names_got.intersection(field_names_expected)`. 😄

ssz/sedes/serializable.py
ChihChengLiang: The intention is to import `tools` when needed since they are not core functions. But moving them to the top looks simpler.

ChihChengLiang: Ah, importing at the top has circular import issue.

ssz/tools/io.py
ChihChengLiang: Maybe `field_names_got.difference(field_names_expected)` is the desired?

ssz/sedes/serializable.py
pipermerriam: I don't think this should be included at all in the `Serializable` API at all.  `library.toolz` is meant to be completely separate from the main library functionality and we should have no problem implementing this functionality in a way that keeps `Serializable` from needing to be aware of it.


```python
# ssz/toolz/dump.py

def to_formatted_dict(value, sedes=None):
    if sedes is None:
        if isinstance(value, Serializable):
            sedes = type(value)
        else:
            raise Error("must provide sedes for non-Serializable"0
    ...  # do the conversion
```

ssz/sedes/serializable.py
```diff
@@ -179,6 +179,17 @@ def __deepcopy__(self, *args):
     def root(self):
         return ssz.hash_tree_root(self)
 
+    def to_formatted_dict(self, int_as="int"):
```
pipermerriam: I've grown to treat magic strings as a code smell (error prone pattern).  I'd advocate for this to either become an `enum.Enum` or better, to pass around an `basic_sedes_formatter` that is responsible for doing the formatting of the basic types.

```python
class DefaultFormatter:
    @staticmethod
    def format_integer(value: int):
        return value
    @staticmethod
    def format_bool(value: bool):
        return value
```

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+)
+from ssz.sedes.serializable import (
+    BaseSerializable,
+    MetaSerializable,
+)
+
+
+class FormattedDictIO:
+    def __init__(self, int_as):
+        self.int_as = int_as
```
pipermerriam: See my comment about using an interface rather than hard coding the various ways things can be formatted.  We have a defined set of basic types so it seems reasonable to just define the interface for dumping/loading and then to pass an instance of that interface into this API.

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+        if self.int_as == "int":
+            return value
+        elif self.int_as == "str":
+            return str(value)
+        elif self.int_as == "hex":
+            return encode_hex(value.to_bytes(size, "little"))
+        else:
+            raise TypeError('Expected "int", "str", or "hex"')
+
+    def parse_value(self, value, sedes=None):
```
pipermerriam: It doesn't appear this function correctly handles the case for `sedes == None` anywhere.

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+            return {
+                field_name: self.parse_value(field_value, sedes.field_name_to_sedes[field_name])
+                for field_name, field_value in value.items()
+            }
+
+        elif isinstance(sedes, BaseSedes):
+            raise Exception("Unreachable: All sedes types have been checked")
+
+        else:
+            raise TypeError("Expected BaseSedes")
```
pipermerriam: It's good to include more information in error messages like this such as what type `sedes` actually was.

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+    UInt,
+    Vector,
+)
+from ssz.sedes.serializable import (
+    BaseSerializable,
+    MetaSerializable,
+)
+
+
+class FormattedDictIO:
```
pipermerriam: I would suggest getting rid of this class entirely as it doesn't appear to be needed, in favor of two functions:

> NOTE: this somewhat hinges on the concept of defining an interface for the dumping/loading of basic values

```python
def dump_value(basic_value_handler, value, sedes):
    ...
def parse_value(basic_value_handler, value, sedes):
    ...
```

These functions can then just recursively walk the individual objects.

ssz/tools/io.py
```diff
@@ -0,0 +1,150 @@
...
+            return tuple(self.parse_value(element, sedes.element_sedes) for element in value)
+
+        elif isinstance(sedes, (ByteList, ByteVector)):
+            return decode_hex(value)
+
+        elif isinstance(sedes, Container):
+            if not isinstance(value, Mapping):
+                raise ValueError(f"Expected mapping, got {type(value)}")
+            field_names_got = set(value.keys())
+            field_names_expected = set(field_name for field_name, _ in sedes.fields)
```
pipermerriam: I'd recommend splitting this logic up into individual functions:

```python
def dump(...):
    if isinstance(sedes, Container):
        return dump_container(...)
    elif ...

def dump_container(...):
    ...

def dump_vector(...):
    ...
```

This should be significantly easier to read and maintain (and to test in isolation)

ssz/toolz/dump.py
```diff
@@ -0,0 +1,107 @@
...
+)
+
+
+def to_formatted_dict(value, sedes=None, basic_sedes_formatter=DefaultFormatter):
+    return dump(value, sedes, basic_sedes_formatter)
+
+
+def dump(value, sedes=None, basic_sedes_formatter=DefaultFormatter):
+    if sedes is None:
+        if isinstance(value, Serializable):
```
pipermerriam: Te way the current code is structured, if someone called `dump(Thing(...), sedes=Thing)` we would end up hitting the final `else` block and throw a `TypeError`.  There should be a test that demonstrates this.  I think this behavior would be surprising.

I would advocate for restructuring this as:

```python
if sedes is None:
    if isinstance(value, Serializable):
        sedes = type(value)
    else:
        raise ...

if isinstance(sedes, Serializable):
    dump_serializable(value, sedes, ...)
elif ...
```

This structure *allows* for you to try and dump a `Serializable` using some other sedes that isn't its class.  If this isn't desirable, then we can tighten the restriction and say that `isinstance(value, Serializable) and sedes is not type(value)` is an error case, though we should be explicit about raising a properly worded error for it to provide good error messages.

ssz/toolz/dump.py
```diff
@@ -0,0 +1,107 @@
...
+
+def dump_integer(value, sedes, basic_sedes_formatter):
+    return basic_sedes_formatter.format_integer(value, sedes)
+
+
+def dump_bytes(value, sedes, basic_sedes_formatter):
+    return basic_sedes_formatter.format_bytes(value, sedes)
+
+
+def dump_list(value, sedes, basic_sedes_formatter):
```
pipermerriam: noting that this code is missing type hints.

ssz/toolz/dump.py
```diff
@@ -0,0 +1,107 @@
...
+    if not isinstance(value, Sequence):
+        raise ValueError(f"Expected Sequence, got {type(value)}")
+    if not len(value) == sedes.length:
+        raise ValueError(f"Expected {sedes.length} elements, got {len(value)}")
+    return tuple(dump(element, sedes.element_sedes, basic_sedes_formatter) for element in value)
+
+
+@to_dict
+def dump_container(value, sedes, basic_sedes_formatter):
+    if not isinstance(value, Mapping):
```
pipermerriam: Noting that this will need to be cleaned up to the new `Container` format that doesn't use field names.

ssz/toolz/formatters.py
```diff
@@ -0,0 +1,32 @@
+from eth_utils import (
+    decode_hex,
+    encode_hex,
+)
+
+
+class DefaultFormatter:
```
pipermerriam: Terminology wise, we might call this a `Codec` and continue using the terms `encode_thing` and `decode_thing`.  It simplifies the naming of things in the dump functions to just be `dump_vector(value, sedes, codec=codec)`

ssz/toolz/dump.py
ChihChengLiang: The `dump(Thing(...), sedes=Thing)` case is handled in 
```
elif isinstance(sedes, MetaSerializable):
        return dump_serializable(value, basic_sedes_formatter)
```
I'll add a test to demenstrate it.

pipermerriam: Ah, I missed the `MetaSerializable` part.  :+1:

pipermerriam: I am still of the opinion that this function would be easier to read if it was done as:

```
def dump():
   if sedes is None:
       # This code only raises errors or sets the `sedes` value.

    ```
    # All of the "dump" code is here, fully separate from the error raising code above.
```

Probably a minor win and entirely possible that I was just reading hastily... do as you please.

ssz/toolz/dump.py
ChihChengLiang: It's a little bit difficult to get the typing right at this stage. Is there any goal for typing we can pursuit now?

ssz/toolz/dump.py
ChihChengLiang: Ah, missed this one, will fix

PR #68, opened 2019-05-01
hash_tree_root cache

### LINKED ISSUE #67: Use `hash_tree_root` as the basic hash of `BaseSerializable`
## What was wrong?

In `BaseSerializable`, it uses `hash(self)` or `hash(tuple(self))` as the basic hash and cache it:
https://github.com/ethereum/py-ssz/blob/9d4bb98bc5d312a34d5188b5a74fb5d9726578a8/ssz/sedes/serializable.py#L134-L152

https://github.com/ethereum/py-ssz/blob/9d4bb98bc5d312a34d5188b5a74fb5d9726578a8/ssz/sedes/serializable.py#L178-L180

But since we use `hash_tree_root` in ssz, we can remove the built-in `hash()`:
1. Cache `self.root` in `py-ssz` level.
2. Use `self.root` in `__eq__` 

## How can it be fixed?

ditto

hwwhww:
## What was wrong?

Fix #67 

## How was it fixed?

1. Add `root_cache` and remove `hash_cache`.
2. Use `self.root == other.root` in `__eq__`.

#### Cute Animal Picture
bird-691816_640

### DIFF
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -132,24 +132,9 @@ class BaseSerializable(collections.Sequence):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return self.__class__ is other.__class__ and hash(self) == hash(other)
+        return self.__class__ is other.__class__ and self.root == other.root
 
-    def __getstate__(self):
-        state = self.__dict__.copy()
-        # The hash() builtin is not stable across processes
-        # (https://docs.python.org/3/reference/datamodel.html#object.__hash__), so we do this here
-        # to ensure pickled instances don't carry the cached hash() as that may cause issues like
-        # https://github.com/ethereum/py-evm/issues/1318
-        state['_hash_cache'] = None
-        return state
-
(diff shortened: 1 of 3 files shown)


ssz/sedes/serializable.py
```diff
@@ -132,24 +132,9 @@ def __len__(self):
...
-        # The hash() builtin is not stable across processes
-        # (https://docs.python.org/3/reference/datamodel.html#object.__hash__), so we do this here
-        # to ensure pickled instances don't carry the cached hash() as that may cause issues like
-        # https://github.com/ethereum/py-evm/issues/1318
-        state['_hash_cache'] = None
-        return state
-
-    _hash_cache = None
-
-    def __hash__(self):
```
jannikluhn: According to the [docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__), two objects that are equal should have the same hash. So we should probably not remove it, but replace it with something like

```Py
return hash(self.__class__) * int.from_bytes(self.root, "little")
```

PR #69, opened 2019-05-02
[Minor] import ssz; ssz.tools.do_something

ChihChengLiang:
## What was wrong?

Y U no `ssz.tools.from_formatted_dict`

## How was it fixed?

`from . import  tools` in `ssz/__init__.py`

#### Cute Animal Picture

image-compressor (1)

### DIFF
--- a/ssz/__init__.py
+++ b/ssz/__init__.py
@@ -1,6 +1,9 @@
 #
 # sedes
 #
+from . import (  # noqa: F401,
+    tools,
+)
 from .codec import (  # noqa: F401
     decode,
     encode,
--- a/tests/tools/test_parsing_and_dumping.py
+++ b/tests/tools/test_parsing_and_dumping.py
@@ -5,6 +5,7 @@ from eth_utils import (
     encode_hex,
 )
 
+import ssz
 from ssz.examples.zoo import (
     Animal,
     Zoo,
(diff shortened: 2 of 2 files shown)


ssz/__init__.py
```diff
@@ -1,6 +1,9 @@
 #
 # sedes
 #
+from . import (  # noqa: F401,
+    tools,
+)
```
pipermerriam: I don't think this is ok.  The idea behind the `library.tools` abstraction is that the main library code never imports it because it might have dependencies that are not installed by default.  We can keep this **if** we have a CI test that does a naive install of the library (without any testing or development dependencies) and verifies that `import ssz` does not error out.

tests/tools/test_parsing_and_dumping.py
```diff
@@ -14,31 +15,26 @@
...
-    DefaultCodec,
-    from_formatted_dict,
-    to_formatted_dict,
-)
 
 
 def test_parsing_and_dumping():
-    json_str = json.dumps(to_formatted_dict(zoo))
-    read_zoo = from_formatted_dict(json.loads(json_str), Zoo)
+    json_str = json.dumps(ssz.tools.to_formatted_dict(zoo))
```
pipermerriam: In general I think this is a pattern we should be avoiding.  It makes grepping for import statements hard and in general goes against the "be explicit" philisophy.  What was wrong with importing the necessary things from `ssz.tools` explicitely?

ssz/__init__.py
ChihChengLiang: Thanks for the review. I was not aware that it reintroduced the dependency back and didn't realize `ssz.tools.to_formatted_dict` is a bad import pattern. Give me a 👍and I'll revert this PR.

PR #71, opened 2019-05-02
Allows a more flexible relationship when comparing classes for `__eq__`

ralexstokes:
Namely, one is a subclass of the other

Following up on @jannikluhn's idea to make class equality more flexible

#### Cute Animal Picture

### DIFF
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -132,7 +132,12 @@ class BaseSerializable(collections.Sequence):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return self.__class__ is other.__class__ and self.root == other.root
+        satisfies_class_relationship = (
+            issubclass(self.__class__, other.__class__) or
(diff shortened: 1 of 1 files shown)


ssz/sedes/serializable.py
```diff
@@ -132,7 +132,12 @@ def __len__(self):
         return len(self._meta.fields)
 
     def __eq__(self, other):
-        return self.__class__ is other.__class__ and self.root == other.root
+        satisfies_class_relationship = (
+            issubclass(self.__class__, other.__class__) or
+            issubclass(other.__class__, self.__class__)
+        )
+        same_root = self.root == other.root
+        return satisfies_class_relationship and same_root
```
pipermerriam: Performance-wise doing so that it's lazy is probably going to be measurable since root calculation is expensive.

```python
if not satisfies_class_relationship:
    return False
else:
    return self.root == other.root
```

ralexstokes: fair

PR #73, opened 2019-06-14
fix containers, direct to field sedes elements, not the pair containing them

protolambda:
## What was wrong?

`fields_sedes` is a series of `(name, sedes)`, not just `sedes`. The code however, expected `sedes` in some places.

Stumbled on it with work on https://github.com/ethereum/eth2.0-specs/pull/1178

## How was it fixed?

Simply unpack the pair, and get the `sedes` from it.

#### Cute Animal Picture

pineapple night owl


protolambda:
Nvm, had incorrect Container instantiation, which resulted in the tuple wrapping of fields.

PR #74, opened 2019-06-14
fix list deserialization

protolambda:
partition manually, to have bytes instead of tuples as BytesIO input

## What was wrong?

`partition` returns a series of tuples. Tuples are not accepted by BytesIO

## How was it fixed?

Manually slice the input. Credits to @pipermerriam for confirming the problem and suggesting a fix.

#### Cute Animal Picture

### DIFF
--- a/ssz/sedes/list.py
+++ b/ssz/sedes/list.py
@@ -12,7 +12,6 @@ from eth_utils import (
 )
 from eth_utils.toolz import (
     cons,
-    partition,
     sliding_window,
 )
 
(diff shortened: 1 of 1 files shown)


ssz/sedes/list.py
```diff
@@ -104,7 +103,8 @@ def _deserialize_stream(self, stream: IO[bytes]) -> Iterable[TDeserialized]:
                     f"element size. data length: {len(data)}  element size: "
                     f"{element_size}"
                 )
-            for segment in partition(element_size, data):
+            for start_idx in range(0, len(data), element_size):
+                segment = data[start_idx: start_idx + element_size]
```
hwwhww: I assume that `partition ` is faster since it's in Cython. If so, can we do:
```python
            for segment in partition(element_size, data):
                segment = list(segment)
```
? /cc @pipermerriam

pipermerriam: I think this is fine for now.  We can attack performance issues once we're a little further along and we know it's a bottleneck.

protolambda: that should be a wrapper in `bytes` then

PR #75, opened 2019-06-19
[POC] Implement fields as descriptors

jannikluhn:
POC of how an alternative fields implementation using Python descriptors instead of a list of fields. This provides a nicer interface and type hints, at least in some cases.

### DIFF
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -1,12 +1,13 @@
 import abc
 import collections
 import copy
-import operator
-import re
+from itertools import (
+    dropwhile,
+    takewhile,
+)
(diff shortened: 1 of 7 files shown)


ssz/sedes/serializable.py
```diff
@@ -40,12 +36,34 @@
 TSerializable = TypeVar("TSerializable", bound="BaseSerializable")
 
 
+class Field(abc.ABC):
```
pipermerriam: If `Field` was done as `class Field(typing.Generic[TValue])` then I think we could get a lot of the type stuff for free on the individual field types without having to implement as many methods.

PR #80, opened 2019-07-12
Rename `hash_tree_root` function to `get_hash_tree_root` and `root` property to `hash_tree_root`

### LINKED ISSUE #79: Avoid collisions on `root` property?
# What is wrong?

If a class inherits from one of our ssz types, then it gets a `root` property that caches the `hash_tree_root` computation.

There is a collision that occurs if this class has another property or even an `ssz` field w/ the name `root`.

This collision caused a bug w/ our `__eq__` instance which compares the `root` property to determine equivalence.

# How to fix it?

If we are just concerned about fixing the `__eq__` instance, then we can try this to modify the property/method resolution order:

```python
  def __eq__(self, other):
        satisfies_class_relationship = (
            issubclass(self.__class__, other.__class__) or
            issubclass(other.__class__, self.__class__)
        )

        if not satisfies_class_relationship:
            return False
        else:
            # current solution:
            # return self.root == super.root
            return super(self.__class__, self).root == super(other.__class__, other).root
```

This seems to avoid the "leaf" `root` property (defined directly on our child class) and instead searches in any of the parent classes for a `root` property -- which then defers to the `hash_tree_root` definition we want.

hwwhww:
## What was wrong?

Fix #79 

## How was it fixed?

- Rename `root` property to `hash_tree_root` for conflict No.1.
- And then found out we need to rename `hash_tree_root` function to `get_hash_tree_root` for conflict No.2.

#### Cute Animal Picture

🐘

### DIFF
--- a/scripts/benchmark/tree_hash.py
+++ b/scripts/benchmark/tree_hash.py
@@ -100,7 +100,7 @@ def prepare_byte_vector_benchmark():
 
     def benchmark():
         for data_item in data:
-            ssz.hash_tree_root(data_item, byte_vector)
+            ssz.get_hash_tree_root(data_item, byte_vector)
 
     return benchmark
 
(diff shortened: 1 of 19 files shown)


ralexstokes:
looks good to me!

PR #83, opened 2019-07-23
Fix List hash_tree_root and merkleize (v0.8.1)

hwwhww:
## What was wrong?

Just realize there is some more work on `List` in #82.

## How was it fixed?
1. Set `chunk_count` as a method for all sedes, to be more like the spec description.
2. Fix `List.get_hash_tree_root()` - it uses `chunk_count = self.max_length` when the element is composite type.

#### Cute Animal Picture

🦉


ssz/sedes/byte_list.py
```diff
@@ -29,8 +29,11 @@ def deserialize(self, data: bytes) -> bytes:
 
     def get_hash_tree_root(self, value: bytes) -> bytes:
         merkle_leaves = pack_bytes(value)
-        merkleized = merkleize(merkle_leaves)
+        merkleized = merkleize(merkle_leaves, limit=self.chunk_count())
         return mix_in_length(merkleized, len(value))
 
+    def chunk_count(self) -> int:
+        pass
```
hwwhww: byte_list is deprecated (#78)

ssz/sedes/boolean.py
```diff
@@ -33,6 +33,9 @@ def deserialize(self, data: bytes) -> bool:
                 f"{encode_hex(data)})",
             )
 
+    def chunk_count(self) -> int:
+        return 1
```
ralexstokes: do we want to explicitly define this here? seems like we can rely on the parent implementation

ssz/sedes/byte_list.py
ralexstokes: shall we just delete this file?

ssz/sedes/byte_vector.py
```diff
@@ -59,6 +59,9 @@ def get_hash_tree_root(self, value: bytes) -> bytes:
         serialized_value = self.serialize(value)
         return merkleize(pack_bytes(serialized_value))
 
+    def chunk_count(self) -> int:
+        return self.length * self.size
+
```
ralexstokes: i'd suggest we delete this file and just define the `bytesN` (which iirc we do use in other places) in (e.g.) the `__init__.py` and just use the `Vector[Byte,N]` construction instead of keep this specialized implementation around.

jannikluhn: A `Vector[Byte, N]` would iterate over the single bytes for serialization/deserialization which is needlessly inefficient, that's why we have this optimized version. As far as I know we haven't benchmarked it though.

hwwhww: I tried to replace it with `Vector(Byte, N)` from our codebase, it's also helpful for dump/parse as `bytes` instead of an array of `Byte()`. Keep the status quo here. :)

ralexstokes: gotcha thanks @jannikluhn -- i'd imagine we could detect if our `Vector[T,N]` has `T == byte` and do the optimizations behind the curtain but i'll also agree status quo works fine for now and may not be worth the maintenance overhead

PR #84, opened 2019-08-12
[Turbo] Rework `BaseSerializable.copy` and `BaseSerializable.__deepcopy__`

hwwhww:
## What was wrong?

Previously on https://github.com/ethereum/trinity/pull/848: `deepcopy` operation is the bottleneck.

## How was it fixed?
1. Rework `copy`:
    - Only need to deep copy if the field is mutable. 
2. Customize `deepcopy`:
    - deepcopy the `self.__dict__.items()` and generate a new object.

### Performance
`247.03 seconds` -> `101.59 seconds` improvement for `test_demo.py`.

### New profile

Note: If this approach is accepted, `pyrlp` could be improved too. It's about 3.5x better for `block.copy()` operation.

#### Cute Animal Picture

cute-animal-pictures-animals-eating-watermelon-010-640x400

### DIFF
--- a/scripts/benchmark/tree_hash.py
+++ b/scripts/benchmark/tree_hash.py
@@ -1,114 +1,216 @@
 import cProfile
+import io
+import pstats
 import random
 import time
 
 import ssz
 from ssz.sedes import (
-    ByteVector,
     List,
     Serializable,
+    Vector,
+    boolean,
     bytes32,
     bytes48,
-    uint32,
+    bytes96,
     uint64,
 )
 
 random.seed(0)
 
-do_profiling = False
+do_profiling = True
+
 TOLERABLE_PERFORMANCE = 15  # Seconds
 
+VALIDATOR_REGISTRY_LIMIT = 2**40  # (= 1,099,511,627,776)
+EPOCHS_PER_HISTORICAL_VECTOR = 2**16  # (= 65,536)
 
-class ValidatorRecord(Serializable):
+FAR_FUTURE_EPOCH = 0
+
+
+class BeaconBlockHeader(ssz.SignedSerializable):
+
+    fields = [
+        ('slot', uint64),
+        ('parent_root', bytes32),
+        ('state_root', bytes32),
+        ('body_root', bytes32),
+        ('signature', bytes96),
+    ]
+
+
+class Eth1Data(ssz.Serializable):
+
+    fields = [
+        ('deposit_root', bytes32),
+        ('deposit_count', uint64),
+        ('block_hash', bytes32),
+    ]
+
+    def __init__(self,
+                 deposit_root,
+                 deposit_count,
+                 block_hash) -> None:
+        super().__init__(
+            deposit_root=deposit_root,
+            deposit_count=deposit_count,
+            block_hash=block_hash,
+        )
+
+
+class Validator(ssz.Serializable):
     fields = [
         ('pubkey', bytes48),
         ('withdrawal_credentials', bytes32),
-        ('randao_commitment', bytes32),
-        ('randao_layers', uint64),
-        ('status', uint64),
-        ('latest_status_change_slot', uint64),
-        ('exit_count', uint64),
-        ('poc_commitment', bytes32),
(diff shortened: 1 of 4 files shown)


ssz/sedes/serializable.py
```diff
@@ -169,21 +170,41 @@ def copy(self, *args, **kwargs):
...
         return self.copy()
 
-    def __deepcopy__(self, *args):
-        return self.copy()
+    def __deepcopy__(self, memodict=None):
+        if memodict is None:
+            memodict = {}
+
+        cls = self.__class__
+        result = cls.__new__(cls)
```
jannikluhn: this looks strange to me. Why can't we do something like this?

```Py
result = cls(**{
    key: copy.deepcopy(value) for key, value in self.to_dict().items()
})
```

(or maybe, while we're at it, even add a classmethod `from_dict`)? Or is calling `__init__` too expensive performance-wise?

ssz/sedes/serializable.py
```diff
@@ -169,21 +170,41 @@ def copy(self, *args, **kwargs):
...
-        unchanged_kwargs = {
-            key: copy.deepcopy(value)
-            for key, value
-            in self.as_dict().items()
-            if key in missing_overrides
-        }
+        unchanged_kwargs = {}
+        for key, value in self.as_dict().items():
+            if key in missing_overrides:
+                is_immutable_field = (
```
jannikluhn: couple of things:
- it's probably better to move this to its own function because it's quite a long expression (e.g. `is_immutable_field_value`)
- shouldn't this check be recursive? So that `((1, 2, 3), (4, 5, 6))` counts as immutable as well.
- this seems to count an empty tuple as mutable as well which is incorrect (even though it probably doesn't matter much regarding performance).

- if we have a separate function for this, we can use a dict comprehension:

```Py
unchanged_kwargs = {
    key: value if is_immutable_field_value(value) else copy.deepcopy(value)
    for key, value in self.as_dict().items()
    if key in missing_overrides
}
```

scripts/benchmark/tree_hash.py
```diff
@@ -1,114 +1,216 @@
...
+def update_tuple_item(tuple_data,
+                      index,
+                      new_value):
+    """
+    Update the ``index``th item of ``tuple_data`` to ``new_value``
+    """
+    return update_tuple_item_with_fn(
+        tuple_data,
+        index,
+        lambda *_: new_value
```
jannikluhn: shouldn't this just be
```suggestion
        lambda _: new_value
```
as we don't have any `*args`?

ssz/sedes/serializable.py
```diff
@@ -169,21 +170,41 @@ def copy(self, *args, **kwargs):
...
+                    isinstance(value, BASE_TYPES) or
```
pipermerriam: it might be faster here to do `if type(value) in BASE_TYPES` as `isinstance` checks can be expensive and we have no need to support subclasses of `int`, `bool`, or `bytes` that I know of.  The same could go for the `isinstance(value, tuple)` check below.

scripts/benchmark/tree_hash.py
ralexstokes: p sure this was copied over from the update i had for the v0.8 spec update...

the idea was to ignore any and all positional inputs to this function (and in this case just return some value to do an unconditional swap). i ran into a case where it was helpful to ignore all positional args but it would be worthwhile to revisit our usage of this idiom to see if we still need it.

a more complete solution is something like `lambda *_, **__: new_value` but we are taking a shortcut in that we don't pass any keyword args here (and expect no one else will either)

hwwhww: Yeah, it was copied from Trinity :p
I opened an issue to track it: https://github.com/ethereum/trinity/issues/911

ssz/sedes/serializable.py
hwwhww: Great idea! <3
Applied in https://github.com/ethereum/py-ssz/pull/84/commits/5dbdb2ba667b9521405c2f609d45d041517ddf04

PR #85, opened 2019-08-13
[Turbo] Cache Merkle leaves

hwwhww:
## What was wrong?

Need to cache Merkle leaves

## How was it fixed?
1. Rework `sedes.base.py`, extract the "basic" classes into `basic.py` to resolve the dependency problem between the Python files.
2. Cache Merkle leaves in `BaseSerializable.cache` object:
    - Two kinds of cache:
        - If `cache` is not assigned, use simple `dict`.   **updated: use `LRU(1024)`**
        - If `cache` is assigned, use the given cache object. The user can choose to use `SSZCache` object or other self-defined `MutableMapping` object.
    - Implement `def get_hash_tree_root_and_leaves(self, value: TSerializable, cache: CacheObj) -> Tuple[Hash32, CacheObj]` for each sedes. It passes `cache` into `merklelize` function.
3. Update `merkleize` function: it first checks if the needed value is in `cache`; if not, `hash(a, b)`.
4. Use serialized result as the cache key <-- which added overhead for more serialization ops

## TODO in other PRs:
1. Add TTL?
2. Clean up `merkleize`
3. Refactor the `serialize()` functions to avoid `has_fixed_size_section_length_cache = hasattr(value, '_fixed_size_section_length_cache')` pattern.

### Performance
Compare to #84

`101.59 seconds` -> `54.28` improvement for test_demo.py

### New profiling
svg

#### Cute Animal Picture

cute-animal-pictures-animals-eating-watermelon-007


ssz/hash.py
```diff
@@ -1,10 +1,12 @@
+import functools
 import hashlib
 
 from eth_typing import (
     Hash32,
 )
 
 
+@functools.lru_cache(maxsize=2**30)
```
pipermerriam: This seems like a very large cache.  Ignoring the input data, I think this is 32 GB worth of hashes....

ssz/sedes/base.py
```diff
@@ -73,6 +76,10 @@ def get_hash_tree_root(self, value: TSerializable) -> bytes:
     def chunk_count(self) -> int:
         pass
 
+    @abstractmethod
+    def get_key(self, value: Any) -> bytes:
+        pass
```
pipermerriam: We recently switched to doing `abstractmethod` this way.

```python
@abstractmethod
def get_key(self, value: Any) -> bytes:
    ...
```

Thoughts?  Intended to help visually separate the `abstractmethod` empty body from an actual empty function body.

ssz/sedes/base.py
```diff
@@ -99,17 +106,30 @@ def get_hash_tree_root(self, value: TSerializable) -> bytes:
         serialized_value = self.serialize(value)
         return merkleize(pack((serialized_value,)))
 
+    def get_hash_tree_root_and_leaves(self, value: TSerializable, db) -> bytes:
+        serialized_value = self.serialize(value)
+        root, db = merkleize(
+            pack((serialized_value,)),
+            db=db,
```
pipermerriam: maybe this would be better named `cache`

ssz/sedes/base.py
```diff
@@ -130,13 +150,21 @@ def serialize(self, value: TSerializable) -> bytes:
         if not len(value):
             return b''
 
-        pairs = self._get_item_sedes_pairs(value)
+        pairs = self._get_item_sedes_pairs(value)  # slow
         element_sedes = tuple(sedes for element, sedes in pairs)
 
-        fixed_size_section_length = _compute_fixed_size_section_length(element_sedes)
+        has_fixed_size_section_length_cache = hasattr(value, '_fixed_size_section_length_cache')
```
pipermerriam: This approach is probably fine, though I think you can get rid of this complexity by relegating this to an `abstractmethod` and letting each implementation determine if it wants to cache the result value or not.

That way this stays a one-liner and each sedes can decide if it wishes to cache the value or not.

ssz/sedes/serializable.py
```diff
@@ -89,9 +92,9 @@ def merge_args_to_kwargs(args, kwargs, arg_names):
 
 
 class BaseSerializable(collections.Sequence):
-    _cached_ssz = None
+    db = None
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, db=None, *args, **kwargs):
```
pipermerriam: I suspect it would be better to put the `db` argument after `*args` so that it's required to be passed in as a keyword argument.

ssz/sedes/serializable.py
```diff
@@ -178,7 +183,22 @@ def copy(self, *args, **kwargs):
...
+        result = type(self)(**all_kwargs)
+
+        if clear_cache:
+            result.clear_cache()
+
+        return result
+
+    def clear_cache(self):
+        if hasattr(self.db, 'reset_cache'):
+            self.db.reset_cache()
```
pipermerriam: `collections.abc.MutableMapping` defines `clear` as the method that does this.  Probably worth using that API.

ssz/sedes/serializable.py
```diff
@@ -162,7 +167,7 @@ def __hash__(self):
             self._hash_cache = hash(self.__class__) * int.from_bytes(self.hash_tree_root, "little")
         return self._hash_cache
 
-    def copy(self, *args, **kwargs):
+    def copy(self, clear_cache=False, *args, **kwargs):
```
pipermerriam: This strikes me as somewhat surprising behavior.  IIUC this cache clearing would extend beyond this individual object  to all objects using this database?  That strikes me as something that shouldn't occur as part of an individual object copy operation but something that would be directly performed on the cache itself.

ssz/utils.py
```diff
@@ -94,6 +96,7 @@ def to_chunks(packed_data: bytes) -> Tuple[bytes, ...]:
         return full_chunks + (last_chunk,)
 
 
+@functools.lru_cache(maxsize=2**30)
```
pipermerriam: These are also very large caches if I'm calculating this correctly...

ssz/hash.py
hwwhww: Yeah! That's a stub now for benchmarking the limitation of this approach for eth2 usecases.

ssz/sedes/base.py
hwwhww: Thanks for the feedback!

I think I have to rework `CompositeSedes.serialize()`, `Vector.serialize()`, `List.serialize()`, and `Cointainer.serialize()` so that we can only check `has_fixed_size_section_length_cache` in the `Cointainer.serialize`.

Since this PR is already too big, will leave it to the next PR.

ssz/sedes/serializable.py
hwwhww: I rework it to [`SSZCache`](https://github.com/ethereum/py-ssz/blob/3000cf09661d6ab67980442a0fda87755dc807fa/ssz/cache/cache.py), it only exists inside `Serializable` object.

The next plan is to make the `Serializable.__init__` accept `db` object, to store the **field values**. (SSZCache is only for Merkle leaves currently)

ssz/cache/cache.py
```diff
@@ -0,0 +1,49 @@
...
+else:
+    MM = MutableMapping
+
+
+class SSZCache(MM):
+    def __init__(self, cache_size: int=2048) -> None:
+        self._cache_size = cache_size
+        self.reset_cache()
+
+    def reset_cache(self) -> None:
```
pipermerriam: According to the [`collections.abc`](https://docs.python.org/3/library/collections.abc.html) classes, this should just be `clear` which is part of the core mutable mapping API.

ssz/cache/cache.py
```diff
@@ -0,0 +1,49 @@
...
+            return super().__contains__(key)
+
+    def __getitem__(self, key: bytes) -> bytes:
+        return self._cached_values[key]
+
+    def __setitem__(self, key: bytes, value: bytes) -> None:
+        self._cached_values[key] = value
+
+    def __delitem__(self, key: bytes) -> None:
+        if key in self._cached_values:
```
pipermerriam: I believe expected behavior from mappings is for `del mapping[key]` to raise a `KeyError` when the value is missing.

ssz/cache/utils.py
```diff
@@ -0,0 +1,59 @@
...
+from ssz.typing import (
+    CacheObj,
+)
+
+
+def get_key(sedes: TSedes, value: Any) -> str:
+    key = _get_key(sedes, value).hex()
+    if len(key) > 0:
+        sedes_name = type(sedes).__name__
+        return sedes_name + _get_key(sedes, value).hex()
```
pipermerriam: Is there a reason you call `_get_key` again rather than just returning `sedes_name + key`?

ssz/sedes/base.py
```diff
@@ -42,143 +26,58 @@ class BaseSedes(ABC, Generic[TSerializable, TDeserialized]):
...
-                              value: Sequence[TSerializable],
-                              ) -> Tuple[Tuple[TSerializable, TSedes], ...]:
-        pass
-
-    def _validate_serializable(self, value: Any) -> None:
-        pass
+    def get_key(self, value: Any) -> bytes:
+        ...
 
-    def serialize(self, value: TSerializable) -> bytes:
```
pipermerriam: Mixing code-reorganization with refactors makes for very difficult review since the diff just shows all new lines.  Was there any logic changed in this method that I should be looking at?

ssz/sedes/container.py
```diff
@@ -162,5 +167,35 @@ def get_hash_tree_root(self, value: Tuple[Any, ...]) -> bytes:
         )
         return merkleize(merkle_leaves)
 
+    def get_hash_tree_root_and_leaves(self, value: Tuple[Any, ...], cache) -> bytes:
```
pipermerriam: Return type here needs to be updated to indicate it returns the `cache` and the `cache` needs a type hint (this should be enforced by CI....)

ssz/sedes/list.py
```diff
@@ -163,6 +184,37 @@ def get_hash_tree_root(self, value: Iterable[TSerializable]) -> bytes:
...
+                )
+            else:
+                merkle_leaves = get_merkle_leaves_without_cache(value, self.element_sedes)
+
+        merkleize_result, cache = merkleize(
+            merkle_leaves,
+            limit=self.chunk_count(),
+            cache=cache,
+        )
+        return mix_in_length(merkleize_result, len(value)), cache
```
pipermerriam: I'm curious why these functions return the `cache`.

ssz/sedes/serializable.py
```diff
@@ -192,17 +208,23 @@ def __deepcopy__(self, memodict=None):
         memodict[id(self)] = result
 
         for k, v in self.__dict__.items():
-            setattr(result, k, copy.deepcopy(v, memodict))
+            if k != 'cache':
+                setattr(result, k, copy.deepcopy(v, memodict))
+
+        setattr(result, 'cache', self.cache)
```
pipermerriam: You should just be able to assign this as `result.cache = self.cache`

ssz/sedes/serializable.py
```diff
@@ -192,17 +208,23 @@ def __deepcopy__(self, memodict=None):
...
+        setattr(result, '_fixed_size_section_length_cache', self._fixed_size_section_length_cache)
```
pipermerriam: same here, no need for `setattr`

ssz/utils.py
```diff
@@ -134,14 +139,19 @@ def hash_layer(child_layer: Sequence[bytes]) -> Tuple[Hash32, ...]:
...
 
     if limit is None:
         limit = chunk_len
 
     if limit == 0:
-        return ZERO_HASHES[0]
+        if cache is None:
+            return ZERO_HASHES[0]
+        else:
+            return ZERO_HASHES[0], cache
```
pipermerriam: Generally I think functions with inconsistent return signatures should be avoided.  They won't play nicely with `mypy` and they tend to require knowledge of the implementation or return value checking to interact with properly.

ssz/sedes/base.py
hwwhww: Sorry! It was moved to `sedes/basic.py` for clear code dependency.
The only changed logic is the `fixed_size_section_length_cache` part:
```python
        has_fixed_size_section_length_cache = hasattr(value, '_fixed_size_section_length_cache')
        if has_fixed_size_section_length_cache:
            if value._fixed_size_section_length_cache is None:
                fixed_size_section_length = _compute_fixed_size_section_length(element_sedes)
                value._fixed_size_section_length_cache = fixed_size_section_length
            else:
                fixed_size_section_length = value._fixed_size_section_length_cache
        else:
            fixed_size_section_length = _compute_fixed_size_section_length(element_sedes)
```
For `hasattr(value, '_fixed_size_section_length_cache')`, there is a commit that I intended to apply your `@abstractmethod` suggestion: https://github.com/hwwhww/py-ssz/commit/1956608a130cba753b9349acd17ae0248c92689b. I think it's still not clear enough for `CompositeSedes` to identify the child class. I'm inclined to rework the `serialize` function inherit flow after this PR.

ssz/utils.py
hwwhww: Yes, totally agreed! This API needs to be refactored a lot and it's in my TODO list.
Workaround for now: returning `Union[Hash32, Tuple[Hash32, CacheObj]]`.

ssz/sedes/list.py
hwwhww: Because I'm not certain we should implement `CacheObj` as a simple mutable `dict`, or we should design an `SSZCache` in the underlayer. 😬Right now it's always:
```python
def  merkleize_related_function_A(..., cache: CacheObj) -> Tuple[Hash32, CacheObj]:
    root, cache = merkleize_related_function_B(..., cache)
    ...

    return final_root, cache
```

A bit tedious.

ssz/cache/utils.py
hwwhww: You are right.

ssz/cache/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def get_key(sedes: TSedes, value: Any) -> str:
+    key = _get_key(sedes, value).hex()
+    if len(key) > 0:
+        sedes_name = type(sedes).__name__
+        return sedes_name + key
+    else:
+        # If the serialized result is empty, use sedes name as the key
+        return key
```
jannikluhn:
```suggestion
        return sedes_name
```

ssz/cache/utils.py
```diff
@@ -0,0 +1,60 @@
...
+)
+from ssz.typing import (
+    CacheObj,
+)
+
+
+def get_key(sedes: TSedes, value: Any) -> str:
+    key = _get_key(sedes, value).hex()
+    if len(key) > 0:
+        sedes_name = type(sedes).__name__
```
jannikluhn: I have the feeling that there should a better way to do this (e.g. define a `merkle_cache_key_prefix` class variable for all sedes objects, and maybe check that there's no name collision using a metaclass). But it's probably good enough for now.

ssz/cache/utils.py
```diff
@@ -0,0 +1,60 @@
...
+@to_tuple
+def get_merkle_leaves_with_cache(value: Any,
+                                 element_sedes: TSedes,
+                                 cache: CacheObj) -> Iterable[Hash32]:
+    """
+    NOTE: cache is mutable
+    """
+    for element in value:
+        key = element_sedes.get_key(element)
+        if key not in cache or len(key) == 0:
```
jannikluhn:
```suggestion
        if key not in cache:
```

`len(key)` should never be `0`, right?

hwwhww:
@pipermerriam Thank you so much for the review!
I'll refactor `merkleize` and rework `serializable <> _fixed_size_section_length_cache` in other PRs.

ssz/cache/utils.py
hwwhww: ahh right, the `len(key) == 0` case was due to the https://github.com/ethereum/py-ssz/pull/85#discussion_r317023361 bug you pointed out.

ssz/cache/cache.py
```diff
@@ -0,0 +1,51 @@
...
+
+from lru import LRU
+
+if TYPE_CHECKING:
+    MM = MutableMapping[bytes, bytes]
+else:
+    MM = MutableMapping
+
+
+class SSZCache(MM):
```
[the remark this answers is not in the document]
hwwhww: I see, now `SSZCache` only serves as an example of how can the upper level (Trinity) implement a new `MutableMapping` subclass instead of simple `LRU()`. e.g., if we want to add TTL.

PR #88, opened 2019-08-26
Rework serialize caches

hwwhww:
## What was wrong?

Avoid some `hasattr` calls.

## How was it fixed?
1. `serialize_cache`
    - Add `serialize_cache` parameter to `Containe.serialize(value, fixed_size_section_length=None, serialize_cache=None) -> bytes`.
2. `fixed_size_section_length`
    - Extract `BasicSedes.get_fixed_size_section_length(value: Sequence[TSerializable], pairs: Optional[Tuple[Tuple[TSerializable, TSedes], ...]]=None)`
    - Refactor & add `fixed_size_section_length` parameter to `BasicSedes.serialize(self, value: TSerializable, fixed_size_section_length=None) -> bytes`. If the given `fixed_size_section_length` is `None`, call  `self.get_fixed_size_section_length(value, pairs)`
3. Apply cache in `Serializable` layer
    - Add `Container.serialize_with_cache(value: TSerializable, fixed_size_section_length_cache: int, serialize_cache: bytes) -> Tuple[bytes, int]`. It returns the serialized result and the `fixed_size_section_length_cache`.
    - In `MetaSerializable.serialize(value)`, it calls `cls._meta.container_sedes.serialize_with_cache(...)` for reusing and storing `value._fixed_size_section_length_cache`. 

Other change: simplify classes by removing `BaseByteSedes`.

Edit:
- Added `class HomogeneousCompositeSedes(CompositeSedes[TSerializable, TDeserialized])` for `List` and `Vector`.
- Added `class NonhomogeneousCompositeSedes(CompositeSedes[TSerializable, TDeserialized])` for `Container`.

#### Cute Animal Picture

tmg-article_tall_wolf

PR #90, opened 2019-08-31
Use as much type information as we have when building a cache key

ralexstokes:
## What was wrong?

I found this while running the `v0.8.3` spec tests. In the case of `CompactCommittee` (https://github.com/ethereum/trinity/blob/master/eth2/beacon/types/compact_committees.py#L12) we have two fields that are each lists of the same size (in this scenario, `4096`).

With the current caching code, we build a cache key that is the type and size => `List4096` so that the cached hash tree root for the second field member was the first field member. Given that the fields are lists over different types,  this caused a mismatch against the spec tests.

## How was it fixed?

Use a more fine grained key for the cache.

#### Cute Animal Picture

### DIFF
--- a/ssz/cache/utils.py
+++ b/ssz/cache/utils.py
@@ -27,7 +27,7 @@ def get_key(sedes: TSedes, value: Any) -> str:
     else:
         # If the serialized result is empty, use sedes name as the key
         if hasattr(sedes, 'element_sedes'):
-            return sedes_name + str(sedes.max_length)
+            return f"{sedes_name}[{sedes.max_length}]({type(sedes.element_sedes).__name__})"
         else:
             return sedes_name
 


ssz/cache/utils.py
```diff
@@ -27,7 +27,7 @@ def get_key(sedes: TSedes, value: Any) -> str:
     else:
         # If the serialized result is empty, use sedes name as the key
         if hasattr(sedes, 'element_sedes'):
-            return sedes_name + str(sedes.max_length)
+            return f"{sedes_name}[{sedes.max_length}]({type(sedes.element_sedes).__name__})"
         else:
```
ralexstokes: we are assuming this is a `List` sedes w/ the access to the `max_length` property -- will all possible structures that bring us here have this property?

hwwhww:
Handling `Serializable`

```python
def get_key(sedes: TSedes, value: Any) -> str:
    key = _get_key(sedes, value).hex()
    if len(key) == 0:
        key = ''

    if hasattr(sedes, 'element_sedes'):
        # If the serialized result is empty, use sedes name as the key
        sedes_name = sedes.__class__.__name__
        return f"{sedes_name}[{sedes.max_length}]({type(sedes.element_sedes).__name__}){key}"
    elif hasattr(sedes, 'cache'):
        field_names = str(sedes._meta.field_names)
        return f"{field_names}{key}"
    else:
        sedes_name = sedes.__class__.__name__
        return f"{sedes_name}{key}"
```

ssz/cache/utils.py
hwwhww: 1. Because `ssz/cache/utils.py` is imported by `ssz/sedes/basic.py`, so it's hacky using `hasattr(sedes, 'element_sedes')` to filter the `List` and `Vector` sedes here. Also, it could be refactored with abstract class: https://github.com/ethereum/py-ssz/pull/88/files#diff-09e4ed3c000952c6e921a58c5c154be5R187https://github.com/ethereum/py-ssz/pull/88/files#diff-09e4ed3c000952c6e921a58c5c154be5R187.
2. Both `Vector` and `List` have `max_length`.

ralexstokes:
Closing in favor of #91.

PR #92, opened 2019-09-03
Fix copy cache

hwwhww:
## What was wrong?

It was removed when I used simple `dict` as the `cache`. Now it should be safe with `LRU` cache.

## How was it fixed?

When `b = a.copy()`, set the `b.cache` to `a.cache`.

#### Cute Animal Picture

cat-2536662_640


hwwhww:
With #91 + #92, `test_demo.py` is now `16.49s`.

what

PR #94, opened 2019-10-17
Add black

### LINKED ISSUE #93: Black
Should we use black in py-ssz? In my opinion it would make sense, even if it's just for the sake of consistency with the other eth2 packages (`eth2` and `py-libp2p`).

cc @hwwhww @ralexstokes @pipermerriam

jannikluhn:
Closes #93 

#### Cute Animal Picture


jannikluhn:
I've also added a [pre-commit](https://pre-commit.com/) config file that people may or may not use to run black, isort, and some other more or less useful checks before committing. Install with `pre-commit install`.

PR #96, opened 2019-10-28
Hashable Structure

jannikluhn:
This implements the base class for hashable structures (lists, vectors, containers). It contains the copy and update logic which should be the majority of the complexity.

Tests aren't really there yet, I'd like to do that after we have benchmarks.

#### Cute Animal Picture

### DIFF
--- a/ssz/abc.py
+++ b/ssz/abc.py
@@ -1,0 +1,136 @@
+from abc import ABC, abstractmethod
+from typing import Any, Generic, Iterable, Iterator, TypeVar, Union
+
+from eth_typing import Hash32
+from pyrsistent.typing import PVector
+
+from ssz.hash_tree import HashTree
+
+TStructure = TypeVar("TStructure")
+TElement = TypeVar("TElement")
+
+
+class HashableStructureAPI(ABC, Generic[TElement]):
+    @property
+    @abstractmethod
+    def raw_root(self) -> Hash32:
+        ...
+
+    #
+    # Element and hash tree access
+    #
+    @property
+    @abstractmethod
+    def elements(self) -> PVector[TElement]:
+        ...
+
+    @property
+    @abstractmethod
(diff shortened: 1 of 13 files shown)


ssz/sedes/byte_vector.py
```diff
@@ -27,6 +28,9 @@ def get_fixed_size(self):
     #
     # Serialization
     #
+    def get_element_sedes(self, index: int) -> BaseSedes:
+        raise NotImplementedError()
```
pipermerriam: It might be nice to have a comment in these cases explaining why this has to be here...

jannikluhn: This will be just temporary. The problem is that bitlists/bitvectors/bytevectors are all technically composite SSZ objects, but don't really have elements that can (or should) be independently serialized. The current HashableStructure just distinguishes between composites and basics for computing the tree root, so it doesn't work for those cases. I still have to implement the method though as it's an abstract method in `BaseCompositeSedes`. I've added a `TODO` comment for now.

ssz/abc.py
```diff
@@ -0,0 +1,136 @@
...
+from eth_typing import Hash32
+from pyrsistent.typing import PVector
+
+from ssz.hash_tree import HashTree
+
+TStructure = TypeVar("TStructure")
+TElement = TypeVar("TElement")
+
+
+class HashableStructureAPI(ABC, Generic[TElement]):
```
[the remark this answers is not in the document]
jannikluhn: I don't think so, because `HashableStructureAPI` is the base class for all hashable structures, so the subclasses already "know" what type they are, so we don't need a type variable for it. `TStructure` is used in some methods of the class, but always bound to `self`, so we don't need to bind it more globally.

[In general, the type annotations in py-ssz are really broken though, and not checked in CI, so don't trust them too much]

PR #97, opened 2019-10-29
Hashable objects

jannikluhn:
Implements HashableList, HashableVector, and HashableContainer, along with some other changes (mostly moving code around and renaming). `HashableVector` and `HashableList` are relatively close to what they'll look like in the end, but `HashableContainer` is just meant for testing and benchmarking for now (most importantly, it doesn't allow element access with `.` notation yet).

#### Cute Animal Picture

animal-animal-photography-baby-orangutan-2083277


jannikluhn:
I made a version of the existing tree hash benchmark that uses the new hashable datastructures to compute the same update. It's very hacky, but it works and the result looks good (first is the existing one, second is the updated one, both with 100k validators):

```
state.hash_tree_root	 0b30f108266c59b2f5ab82d47408667b9cff6e7099cb5a2a6e17001ed4f971c6
state.hash_tree_root	 0b30f108266c59b2f5ab82d47408667b9cff6e7099cb5a2a6e17001ed4f971c6
----- start -----
updated	 fd6a564710f96c042762a7af3819f042447c818c257b1ce8f3a878070db85980
----- end -----
state: 7.4242s
----- start -----
updated	 fd6a564710f96c042762a7af3819f042447c818c257b1ce8f3a878070db85980
----- end -----
state_hashable: 0.0078s
```

Note that the timing does not include the original setup (i.e. calculation of the hash tree in the beginning), just the update. But that's the whole idea, most of the time we update only a small part of the tree.

hwwhww:
@jannikluhn 

(I think) since we cached `hash_eth2` function, so you make sure that the second benchmark is running on a clean environment. I got `state_hashable: 0.0050s` when I first time run `python scripts/benchmark/tree_hash.py`, and got `0.0128s` after I commented out the old Serializable benchmark -- still good though! 🎉

ssz/hashable_container.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC
+from typing import Any, Dict
+
+from eth_typing import Hash32
+
+from ssz.hashable_structure import BaseHashableStructure
+from ssz.sedes import Container
+
+
+class HashableContainer(BaseHashableStructure, ABC):
```
pipermerriam: `ABC` doesn't appear to be needed here (unless I'm missing something)

ssz/hashable_vector.py
```diff
@@ -0,0 +1,21 @@
...
+from pyrsistent import pvector
+
+from ssz.hashable_structure import BaseHashableStructure
+from ssz.sedes import BaseSedes, Vector
+
+
+class HashableVector(BaseHashableStructure):
+    @classmethod
+    def from_iterable(
+        cls: "Type[HashableVector]", iterable: Iterable, element_sedes: BaseSedes
```
pipermerriam: I don't think you need the `Type[HashableVector]` here but maybe I'm mistaken.

PR #104, opened 2019-12-20
Fix deprecation warning

cburgdorf:
## What was wrong?

A deprecated import is causing a warning downstream in Trinity.

```
/home/cburgdorf/Documents/hacking/ef/trinity/venv/lib/python3.7/site-packages/ssz/sedes/serializable.py:70: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class BaseSerializable(collections.Sequence):
```

## How was it fixed?

Import `Sequence` from `collections.abc` rather than `collections`

#### Cute Animal Picture

### DIFF
--- a/ssz/sedes/serializable.py
+++ b/ssz/sedes/serializable.py
@@ -1,5 +1,5 @@
 import abc
-import collections
+import collections.abc
 import copy
 import operator
 import re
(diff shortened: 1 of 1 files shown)


ssz/sedes/serializable.py
```diff
@@ -1,5 +1,5 @@
 import abc
-import collections
+import collections.abc
```
cburgdorf: We are doing this to avoid naming conflicts with `Sequence` that we import from `typing`.

Issue #111, opened 2020-05-06
An empty bitvector is illegal


booleanfunction:
In the [eth2.0-specs](https://github.com/ethereum/eth2.0-specs) it specifies within [simple-serialise.md](https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/simple-serialize.md) (Illegal types section) reads:

> Empty vector types (Vector[type, 0], Bitvector[0]) are illegal.

However in bitvector.py an error is raised only if `bit_count` is negative. An error should be raised if `bit_count` <= 0.

https://github.com/ethereum/py-ssz/blob/651471374ce57627fa4645c6b3e997858794122f/ssz/sedes/bitvector.py#L19-L22

This can be contrast to vector.py where the error is correctly raised if `length` <= 0.

https://github.com/ethereum/py-ssz/blob/651471374ce57627fa4645c6b3e997858794122f/ssz/sedes/vector.py#L30-L32

pipermerriam:
If you'd like to open a pull request fixing this discrepancy I'd be happy to get it reviewed and merged.

carver:
Closed by #112

PR #112, opened 2020-05-07
Prevent bitvectors of length 0

booleanfunction:
## What was wrong?

According to the spec, an empty bitvector is illegal and so an error should be raised if bit_count <=0

Issue #111

## How was it fixed?

The bound on bit_count was adjusted so that an error is raised if it is zero or negative

#### Cute Animal Picture


booleanfunction:
> Would you be willing to write a small test that validates that instantiation of this class with a `bit_count == 0` raises a `TypeError`?

I have added a new test file as requested :)

PR #119, opened 2022-05-24
Encode a byte list more naturally

### LINKED ISSUE #118: Encode a Vector or List of Byte, from bytes
### Code that produced the error

```py
import ssz
from ssz.sedes import Byte, List

List(List(Byte(), 255), 255)
ssz.encode([b'hi', b'bye'], sedes)
```

(Note, this may seem silly when the encoding is trivial
### Full error output

```sh
    def serialize(self, value: bytes) -> bytes:
>       if len(value) != 1:
E       TypeError: object of type 'int' has no len()

self       = <ssz.sedes.byte.Byte object at 0x7f77eee6c040>
value      = 2

venv-py3.8/lib/python3.8/site-packages/ssz/sedes/byte.py:14: TypeError

```

### Expected Result

You can achieve the expected result by using this workaround:

```sh
>>> ssz.encode([[b'h', b'i'], [b'b', b'y', b'e']], sedes)
b'\x08\x00\x00\x00\n\x00\x00\x00hibye'
```

### Environment

Using python 3.8, and `ssz==0.2.4`

## How can it be fixed?

Maybe in `ssz.sedes.byte`, if the value is an `int` and `0 <= value < 256`, then encode it as a byte.

carver:
## What was wrong?

Fix #118 

## How was it fixed?

Add a `ByteList` class that more efficiently encodes a variable-length series of bytes. It's also more convenient than sending in an iterable of single bytes. (since if you just send a `bytes` in python, it will yield each element as an `int` instead, which fails on a `List(Byte)`.

Generally, follow the same pattern as `ByteVector`. Add tests to make sure that `Byte(List)` is equivalent to `ByteList()`.

Skipped `.get_hash_tree_root_and_leaves()` implementation for now, for lack of context.

TODO
- [x] write test
- [x] build docs successfully
- [x] Move to another issue: `ByteList.get_hash_tree_root_and_leaves()`
- [x] Move to another issue: modernize the template, it's very old
- [x] Move to another issue: Upgrade black/isort and autoformat, including setup.py (started in another local branch already)

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,31 +39,31 @@ jobs:
   doctest:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: cimg/python:3.6
         environment:
           TOXENV: doctest
   lint:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: cimg/python:3.6
         environment:
           TOXENV: lint
   py36:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: cimg/python:3.6
         environment:
           TOXENV: py36-core
   yaml:
     <<: *common
     docker:
-    - image: circleci/python:3.6
+    - image: cimg/python:3.6
       environment:
         TOXENV: yaml
   benchmark:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: cimg/python:3.6
         environment:
           TOXENV: benchmark
 workflows:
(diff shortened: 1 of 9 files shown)


carver:
Also see update to #118, this might be the wrong direction.

carver:
I'm blocked on EF operations to fix the missing deploy key. Circle CI is not pulling the source, so can't run the tests.

carver:
This seems roughly identical to `ByteVector`, so I'll ask @jannikluhn to take a look. But also just say if you don't have the time or interest to review, and I'll find someone else. 😄

carver:
Ok, one more person to try. @hwwhww Do you want to take a look?

If no one is excited to maintain this... I may just rage-merge if I don't hear back for a while :)

jannikluhn (approved):
Sorry for the late review. I haven't looked at SSZ or this library in particular in a while, but the PR seems to make a lot of sense, the implementation LGTM, and the animal picture is very cute.

PR #141, opened 2024-11-20
remove pyrsistent upper pin

### LINKED ISSUE #140: Support pyrsistent >= 0.20.0
### What feature should we add?

Add support for pyrsistent 0.20.0 which includes official support for Python 3.12. See https://github.com/tobgu/pyrsistent/blob/master/CHANGES.txt

pacrob:
### What was wrong?

`pyrsistent` dep is has an upper pin that only officially supports up to py38

Related to Issue #140
Closes #140

### How was it fixed?

Remove upper pin on `pyrsistent` dep.

The upper pin exists because we make use of an internal `transform` function, but it hasn't changed for years and the maintainer doesn't expect it to - https://github.com/tobgu/pyrsistent/issues/180

We already have a test for our usage of `transform` in `test_hashable_structures`, so I've added a test to make sure the source code hasn't changed.

### Todo:

- [x] Clean up commit history
- [x] Add entry to the [release notes](https://github.com/ethereum/py-ssz/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,9 +54,7 @@ setup(
     install_requires=[
         "eth-utils>=2",
         "lru-dict>=1.1.6",
-        # When updating to a newer version of pyrsistent, please check that the interface
-        # `transform` expects has not changed (see https://github.com/tobgu/pyrsistent/issues/180)
-        "pyrsistent>=0.18.0",
+        "pyrsistent>=0.16.0",
     ],
     python_requires=">=3.8, <4",
     extras_require=extras_require,
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         "lru-dict>=1.1.6",
         # When updating to a newer version of pyrsistent, please check that the interface
         # `transform` expects has not changed (see https://github.com/tobgu/pyrsistent/issues/180)
-        "pyrsistent>=0.16.0,<0.17",
+        "pyrsistent>=0.18.0",
     ],
     python_requires=">=3.8, <4",
     extras_require=extras_require,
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -56,7 +56,7 @@
         "lru-dict>=1.1.6",
         # When updating to a newer version of pyrsistent, please check that the interface
         # `transform` expects has not changed (see https://github.com/tobgu/pyrsistent/issues/180)
-        "pyrsistent>=0.16.0,<0.17",
```
kclowes, line 59: Could we keep this as `>=0.16.0` or does it break tests?

pacrob: It doesn't break anything when installing with a supported python version, because it'll never be installed. I've just been going with the lowest version that supports our lowest version of python, and 0.16.0 stops at py37 - https://pypi.org/project/pyrsistent/0.16.1/

kclowes: Thinking about this more, I think I'd prefer to keep it at `>=0.16.0` so if someone is using this with an older version of python they still can. Unless we have other breaking changes queued up too?

setup.py
```diff
@@ -56,7 +56,7 @@
         "lru-dict>=1.1.6",
         # When updating to a newer version of pyrsistent, please check that the interface
```
reedsa, line 57: Since this doesnt seem like it will change ever again, should we remove this comment altogether?

pacrob: Removed!

setup.py
pacrob: Leaving at `>=0.16.0`. Next breaking will likely be dropping py38, but shouldn't hurt anything.

kclowes (approved):

fselmo (commented):
lgtm

PR #142, opened 2025-01-17
Upgrade template

pacrob:
### What happen?

Merged template, adding py313 support and replacing `bumpversion` with `bump-my-version`.

`hypothesis` was pinned too low for py313 tests to pass, so bumped it up to `>=6.22.0,<6.108.7`, the same constraints from the recent `eth-abi` update - https://github.com/ethereum/eth-abi/pull/241

### Todo:

- [x] Clean up commit history
- [x] Add entry to the [release notes](https://github.com/ethereum/py-ssz/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image


setup.py
```diff
@@ -7,20 +7,22 @@
 extras_require = {
     "dev": [
         "build>=0.9.0",
-        "bumpversion>=0.5.3",
+        "bump_my_version>=0.19.0",
         "ipython",
+        "mypy==1.10.0",
```
kclowes, line 12: If mypy is disabled, maybe we should exclude it from here for now?

kclowes (approved):
Sweet! I had one question about adding the mypy dep, and I noticed that when I ran `make docs` I had a couple untracked, autogenerated `rst` files named `docs/tests.core.*` that I think should not be generated. Looks good otherwise!

fselmo (approved):
Agreed on leaving out `mypy` for now. Looks good otherwise 👍🏼

setup.py
pacrob: removed!

Issue #137, opened 2024-03-25
rehab consensus spec tests


pacrob:
The current `yaml` tests were being run against:
https://github.com/ethereum/eth2.0-tests/

which were effectively turned off here:
https://github.com/ethereum/py-ssz/pull/85/

There are newer, currently maintained tests here:
https://github.com/ethereum/consensus-spec-tests

We should find relevant tests and make sure we pass them.

PR #143, opened 2025-01-22
add mypy to deps, enable lint type check, first round of fixes

pacrob:
### What was wrong?

Needs updated typing.

### How was it fixed?

Set mypy to run with linting tools. First round of adding typing.

In the interest of keeping PRs reviewable, I'll do smaller ones and merge to the `add-mypy-typing` branch off of `main`.

### Todo:

- [x] Clean up commit history
- [ ] Add or update documentation related to these changes
- [ ] Add entry to the [release notes](https://github.com/ethereum/py-ssz/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,14 +276,14 @@ jobs:
   yaml:
     <<: *common
     docker:
-      - image: cimg/python:3.8
+      - image: cimg/python:3.13
         environment:
           TOXENV: yaml
 
   benchmark:
     <<: *common
     docker:
-      - image: cimg/python:3.8
+      - image: cimg/python:3.13
         environment:
           TOXENV: benchmark
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,7 +47,8 @@ repos:
     hooks:
     -   id: mypy-local
         name: run mypy with all dev dependencies present
-        entry: python -m mypy -p ssz
+        # entry: python -m mypy -p ssz
+        entry: python -m mypy --follow-imports=skip ssz/cache
         language: system
         always_run: true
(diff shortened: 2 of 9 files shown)


reedsa (commented):
Had a look, the typings all look fine 👍 Will approve once lint passes 😸

pacrob:
> Tests won't be passing for each PR

fselmo:
> I'm doing smaller PRs with a bit of typing added in each to make reviewing easier.

I think in the past we have turned on certain noqa or certain typing checks here and there and had them pass with each PR. If we can do that here, it would make things a lot easier to review for actually passing lint. Is there any way we can go about it that way? If it's a big ask then maybe we can re-assess.

.circleci/config.yml
```diff
@@ -276,14 +276,14 @@ jobs:
   yaml:
     <<: *common
     docker:
-      - image: cimg/python:3.8
+      - image: cimg/python:3.13
```
fselmo, line 279: Curious on why this change. Also `yaml` is commented out in the test runs below (L313). I might be missing some context here, but should this be turned on?

pacrob: yaml tests were turned off a while ago, as noted here - https://github.com/ethereum/py-ssz/issues/137

I recall there being an issue with `ruaml.yaml` dep too, but not specifically what.

I just updated it there because I was doing the same for the benchmark tests.

fselmo: Thanks for the background. sgtm, glad we're tracking that with an issue at least 👍🏼

fselmo (approved):
lgtm 👍🏼

PR #144, opened 2025-02-05
Turn on type checking for more files

pacrob:
### What was wrong?

Add types and type checking for:

```
ssz/cache
ssz/sedes/__init__.py
ssz/sedes/base.py
ssz/sedes/basic.py
ssz/utils.py
ssz/constants.py
ssz/abc.py
```

- [x] Clean up commit history
- [ ] Add entry to the [release notes](https://github.com/ethereum/py-ssz/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/ssz/sedes/base.py
+++ b/ssz/sedes/base.py
@@ -7,6 +7,7 @@ from typing import (
     Generic,
     Optional,
     Tuple,
+    TypeVar,
 )
 
 from eth_typing import (
@@ -15,9 +16,10 @@ from eth_typing import (
 
 from ssz.typing import (
     CacheObj,
-    TDeserialized,
-    TSerializable,
 )
+
+TSerializable = TypeVar("TSerializable")
+TDeserialized = TypeVar("TDeserialized")
 
 
 class BaseSedes(ABC, Generic[TSerializable, TDeserialized]):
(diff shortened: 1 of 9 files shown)


ssz/utils.py
```diff
@@ -145,7 +145,7 @@ def _get_merkleized_result(
     max_depth: int,
     cache: CacheObj,
 ) -> Tuple[Hash32, CacheObj]:
-    merkleized_result_per_layers = [None for _ in range(max_depth + 1)]
+    merkleized_result_per_layers = [b"" for _ in range(max_depth + 1)]
```
pacrob, line 148: change is to satisfy the line `key = merkleized_result_per_layers[layer] + node` below. `merkleized_result_per_layers[layer]` is never actually `None` when that line is reached, but setting the list full of `b""` instead has the same effect and satisfies mypy.

ssz/utils.py
```diff
@@ -124,7 +124,7 @@ def get_next_power_of_two(value: int) -> int:
     if value <= 0:
         return 1
     else:
-        return 2 ** (value - 1).bit_length()
+        return int(2 ** (value - 1).bit_length())
```
pacrob, line 127: `**` will only return non-int if `value` is < 0, which we check for, but mypy doesn't understand that.

ssz/sedes/base.py
```diff
@@ -15,10 +16,11 @@
 
 from ssz.typing import (
     CacheObj,
-    TDeserialized,
-    TSerializable,
 )
 
+TSerializable = TypeVar("TSerializable")
+TDeserialized = TypeVar("TDeserialized")
```
pacrob, line 22: These need to be created here rather than imported from typing due how `Generic` works.

reedsa (approved):
:shipit:

